PR #76, opened 2018-08-21
object based segmentation metrics

cpavelchek:


msschwartz21:
Commit fbeb0be7710d90dd3281093e4cce7e0d4c233533 needs to be deleted or reverted to remove the jupyter lab change. I'm not sure how to safely do that on my own.

willgraf:
> Commit [fbeb0be](https://github.com/vanvalenlab/deepcell-tf/commit/fbeb0be7710d90dd3281093e4cce7e0d4c233533) needs to be deleted or reverted to remove the jupyter lab change. I'm not sure how to safely do that on my own.

In the `Dockerfile`, you can just remove the `ENTRYPOINT` line (it will then use the `ENTRYPOINT` defined in the base image).  Also, in `requirements.txt` you can replace lines 7 and 8 with `jupyter>=1.0.0,<2` which is the old jupyter requirement.

msschwartz21:
I did not write tests for the edge cases in `stats_object_based` because the function is deprecated and will be deleted in the next pull request.

Issue #107, opened 2018-11-08
Deepcell Should support CPU and GPU images


willgraf:
Just as tensorflow as images `tensorflow/tensorflow` and `tensorflow/tensorflow-gpu`, deepcell should also support these options for users that may not have access to GPU machines.

willgraf:
Duplicate of #105

willgraf:
One tricky aspect to this is the `tensorflow` vs `tensorflow-gpu` python dependency.  If `tensorflow` is installed via pip, it will cause the container to lose access to the GPU.  however, if `tensorflow-gpu` is installed on a machine without GPU support, we will get CUDA errors.

Somehow it would be great to use `sed` to remove `tensorflow` from the requirements.txt list, but leave it in there by default for users who want to pip install without docker.

willgraf:
Now that travis-ci is automatically building images and pushing them to dockerhub, it seems very possible to build a gpu/cpu version each time.

`sed` still seems like a very good idea.

Issue #105, opened 2018-10-31
cannot install deepcell on a non-GPU machine


willgraf:

willgraf:
This issue is really due to the fact that the docker base image expects to be installed on a GPU machine.  If it is installed on a non-GPU enabled machine, it will fail as it is expecting certain resources on start-up.

I think this may best be resolved in combination with #107 by building several `deepcell` images as they have with tensorflow.

Different build for `deepcell-tf:tag` and `deepcell-tf:tag-gpu`, allowing users to use whichever best suits their needs

willgraf:
This should be easily achievable by updating the base docker image.
However, the `requirements.txt` specifically calls out `tensorflow-gpu`.  How can we make this `requirements.txt` flexible?

willgraf:
Closing as duplicate of #107 which has more information and a more specific call-out.

PR #118, opened 2018-11-28
Use tensorflow/tensorflow as base image

### LINKED ISSUE #115: Remove dependency on NGC Docker base image
Many users will not be able to authenticate with NGC in order to pull the image.  By changing the base image, we can help all users with this issue.

README should be updated as well to indicate that this is no longer an issue.

willgraf:
Deprecating the nvidia base image, as users will need to make an account in order to use this.  Additionally, `tensorflow/tensorflow` has base images that compile for both CPU and GPU, which we can leverage in creating builds for `deepcell` for both CPU and GPU as well.  This base image is a required step in getting CPU and GPU builds, but these will likely need to be automatically generated, or the user will be told what version to use.

Fixes #115 
Related to #107 
Related to #105

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
-# Use the nvidia tensorflow:18.08-py3 image as the parent image
-FROM nvcr.io/nvidia/tensorflow:18.08-py3
+# Use tensorflow/tensorflow as the base image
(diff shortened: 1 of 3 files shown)


willgraf:
Ran through examples of each type of notebook, all continue to function as expected with the different base image.

PR #131, opened 2019-01-02
image_generators.py inherits from keras-preprocessing

### LINKED ISSUE #102: image_generators.py not comminicating with notebooks
Love everything here and would really like to get it running on my system (perhaps i should just use the docker or go back to an older version of tensorflow / keras.preprocessing). Alas I'm new to this.  

I'm just getting errors with importing specific packages that have recently been changed. e.g. 

from tensorflow.python.keras.preprocessing.image import apply_transform
from tensorflow.python.keras.preprocessing.image import flip_axis
from keras_maskrcnn.preprocessing.generator import Generator as _MaskRCNNGenerator

I'm trying to make a workaround for them in my clone of deepcell-tf but am new to tf. Perhaps you are already aware of these issues with newer versions of tf / keras? Or some advice on how to get it running.

### LINKED ISSUE #111: New versions of tensorflow are missing functions from keras_preprocessing
tensorflow gets most of it's functions in keras.preprocessing.image from another repository `keras_preprocessing`.  In the new versions, tensorflow has removed several functions required to build new `Iterator` classes, such as `apply_transform` and `flip_axis`.

These functions should be directly imported from keras_preprocessing.  Additionally, there are several functions in tensorflow that are directly imported from `keras_preprocessing`, without any additional modification or documentation.  I am worried these may also become "deprecated", and perhaps should be directly imported from `keras_preprocessing` as well.  This library will also give us direct access to `transform_matrix_offset_center`.

This is a direct cause of #102, but only for the tensorflow imports.

willgraf:
As of tensorflow version 1.11.0, `tf.keras` updated their ImageDataGenerators to use the `keras-preprocessing` ImageDataGenerator classes.  This is incompatible with the current versions of ImageDataGenerators (the `apply_transform()` has been moved into a class method, `random_transform()` is totally different).

`image_generators.py` has been refactored to support this new parent class, and some helper functions for the old build have been removed.  The ImageDataGenerator class is still imported from `tf.keras`, however, if the imported class does not have the `apply_transform` attribute (it was imported from `tf.keras` version < 1.11.0), then the class is inherited directly from keras-preprocessing.

If it is determined that `deepcell-tf` will only support tensorflow versions after 1.10.0, then we can remove `keras-preprocessing` and its imports from the module and use native tf.keras.

This fixes #111 and fixes #102


willgraf:
Kernel appears to die during `fit_generator` while running notebooks.
Not sure why there is this memory issue, but this is a blocking issue.

PR #177, opened 2019-08-02
update TF base image

MekWarrior:
Updates tensorflow base image and removes an error in the requirements file that invalidates the build argument.


willgraf:
I tried to run a couple of notebooks but it looks like we don't have access to the GPU?
Maybe `train_utils.count_gpus` is broken only, but something is definitely wrong.

PR #180, opened 2019-08-06
Bugfix: export_model bugfix for list output/input

### LINKED ISSUE #167: Export model error
https://github.com/vanvalenlab/deepcell-tf/blob/772a3ecd67da7294dc4c3f4383c3afcc588b5478/deepcell/utils/export_utils.py#L71

I think that the line of code above needs to be changed to `keras_model.output` in order to work correctly. I noticed this error while exporting the Retinanet model so it's possible that the issue was specific to me.

willgraf:
Fixes #167 

Previously we had a check:

```
if type(keras.model.input) is list:
   output = keras.model.output[-1]
```

This should be:

```
if type(keras.model.outut) is list:
   output = keras.model.output[-1]
```


msschwartz21:
This PR may require one more fix in order to enable tracking model export.             [`            predictions.append(self._single_request(stub, request)[-1])`](https://github.com/vanvalenlab/kiosk-redis-consumer/blob/239094e2d75a6cb1c4c03e6ced99249c71140071/redis_consumer/grpc_clients.py#L348)
I think that we need to avoid duplicating the [-1] in this line of code with `output = model.output[-1]`

PR #213, opened 2019-09-03
Add complete cell type datasets for phase and fluorescent cytoplasm

msschwartz21:


msschwartz21:
Not ready to merge yet because the all fluorescent dataset needs to be updated.

msschwartz21:
Local testing in a notebook verified the start of downloads for the three datasets that were updated. Ready to merge.

PR #230, opened 2019-09-22
Update Scale and Label Detection models to use mobilenetv2 backbone.

willgraf:
Updated ScaleDetectionModel architecture and added trained model files for both mobilenet and mobilenet2 for ScaleDetection and LabelDetection.


willgraf:
Currently, the backbones work fine and automatically download the mobilenetv2 weights file.

However, the architecture does not allow for data of different shapes.  for example, the output shape of `(1, 216, 216, 1)` data is `(1, 3)` for label detection, but is `(1, 48)` for data of `(1, 512, 512, 1)`.

Ideally, the architecture should be flexible to the differing input shapes.

Also, is there a real reason to support BOTH mobilenet and mobilenetv2?

willgraf:
The jupyter notebook has been re-run for mobilenetv2 and has been added to the PR.  Removed mobilenet backbone, as it is strictly worse than mobilnetv2 for these use-cases.  Also, it does not reduce coverage 👍

PR #231, opened 2019-09-23
Use keras-applications to standardize backbones

### LINKED ISSUE #173: object detection with P2 layer of RetinaNet with resnet50 backbone fails in tensorflow 1.11
object detection using the P2 layer of RetinaNet with resnet50 backbone fails in tensorflow 1.11.0 (when training on simple test data, the training runs but results are poor). This is fixed starting from tensorflow 1.12 (verified on 1.12.0, 1.14.0).

willgraf:
`tf.keras.applications` tends to vary across versions of tensorflow.  By using `keras-applications` as a direct dependency, we can build the backbones without depending on the particular version of tensorflow.

---

Fixes #173

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -17,4 +17,3 @@ omit =
     deepcell/utils/testing_utils.py
     deepcell/datasets/*
     deepcell/notebooks/*
-    deepcell/tracking.py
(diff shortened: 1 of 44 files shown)


willgraf:
Having issues with testing python2.7 on travis.  Will re-open this when it is resolved.

PR #250, opened 2019-10-23
Remove weights from ImageNormalization layers.

### LINKED ISSUE #249: ValueError when running 3D Watershed
Hi,

I am running **3D Watershed - Fully Convolutional.ipynb** and get the following error when run the code:

**ValueError: Cannot feed value of shape (3, 61, 61, 3, 1) for Tensor 'image_normalization3d_14/Placeholder:0', which has shape '(15, 61, 61, 3, 1)'**

I trained the model, saved it, and when I try to run the model (as follows),  it doesn't like to accept an input shape that is different from training (which is weird!)

**run_fgbg_model = model_zoo.bn_feature_net_skip_3D(receptive_field=receptive_field,
    n_features=2,
    n_frames=frames_per_batch,
    n_skips=n_skips,
    n_conv_filters=32,
    n_dense_filters=128,
    input_shape=tuple(X_test.shape[1:]),
    multires=False,
    last_only=False,
    norm_method=norm_method)
run_fgbg_model.load_weights(fgbg_weights_file)**

willgraf:
For the `ImageNormalization` layers, adding a weight for the kernel can cause errors when loading the weights with a different input shape. By removing the `self.add_weight` call, we can prevent this issue.  Instead of weights, we can just use `self.kernel = W` directly.

---

Resolves #249

### DIFF
--- a/deepcell/layers/normalization.py
+++ b/deepcell/layers/normalization.py
@@ -104,18 +104,19 @@ class ImageNormalization2D(Layer):
         self.input_spec = InputSpec(ndim=4, axes={channel_axis: input_dim})
 
         kernel_shape = (self.filter_size, self.filter_size, input_dim, 1)
-        self.kernel = self.add_weight(
-            name='kernel',
-            shape=kernel_shape,
-            initializer=self.kernel_initializer,
-            regularizer=self.kernel_regularizer,
-            constraint=self.kernel_constraint,
-            trainable=False,
-            dtype=self.dtype)
+        # self.kernel = self.add_weight(
+        #     name='kernel',
+        #     shape=kernel_shape,
+        #     initializer=self.kernel_initializer,
+        #     regularizer=self.kernel_regularizer,
+        #     constraint=self.kernel_constraint,
+        #     trainable=False,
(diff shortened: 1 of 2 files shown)


aziabari:
**I just tested your updated code and the problem still persists.
My training data are 128x128x128.
I have used frames_per_batch=3

I have 4 test volumes (X_test.shape = (4, 128, 128, 128, 1))
When I want to load the model for testing I ran the following code,**

run_fgbg_model = model_zoo.bn_feature_net_skip_3D(
receptive_field=receptive_field,
n_features=2,
n_frames=frames_per_batch,
n_skips=n_skips,
n_conv_filters=32,
n_dense_filters=128,
input_shape=tuple(X_test.shape[1:]),
multires=False,
last_only=False,
norm_method=norm_method)
run_fgbg_model.load_weights(fgbg_weights_file)

**And I get the following error:**

ValueError: Cannot feed value of shape (3, 29, 29, 1, 1) for Tensor 'image_normalization3d_25/Placeholder:0', which has shape '(128, 29, 29, 1, 1)

Please also look at my comments in #249

PR #282, opened 2019-12-13
Modified featurenet to include temporal features that can be residuals

circuit1:
- Temporal features include `Conv3D` (previous standard), `None`, `ConvGRU2D` and `ConvLSTM2D`
- Added new parameters including:
1. `temporal` - the mode of temporal feature, which is either None, `'conv'`, `'gru'` or `'lstm'`
2. `residual` - whether the temporal feature is a residual or not (`True` or `False`). If it is a residual, the input into the next layer of the net is the previous feature + temporal feature. Otherwise, it input is simply the temporal feature.
3. `temporal_kernel_size` - the size of the kernel used in the convolutions for temporal features. The default is `3`, and the layers will automatically expand `temporal_kernel_size` to a kernel of `(temporal_kernel_size, temporal_kernel_size)`

- Modified `featurenet_test.py` to include test cases for the new features and to include parameters for existing test cases

- Ran `python3 -m pytest --cov=deepcell --pep8 deepcell/model_zoo/featurenet_test.py` and got `83 passed, 1 skipped, 13 warnings in 179.78 seconds`


deepcell/model_zoo/featurenet.py
```diff
@@ -446,7 +450,36 @@ def bn_feature_net_3D(receptive_field=61,
...
+        elif mode == 'lstm':
+            x = ConvLSTM2D(filters=n_filters, kernel_size=temporal_kernel_size,
+                           padding='same', kernel_initializer=init, activation='relu',
+                           kernel_regularizer=l2(reg), return_sequences=True)(feature)
+        elif mode == 'gru':
+            x = ConvGRU2D(filters=n_filters, kernel_size=temporal_kernel_size,
+                          padding='same', kernel_initializer=init, activation='relu',
+                          kernel_regularizer=l2(reg), return_sequences=True)(feature)
+        else:
+            return feature
```
willgraf: Should this `raise ValueError`? I am considering the case of someone using "ltsm" as an input.

deepcell/model_zoo/featurenet.py
```diff
@@ -446,7 +450,36 @@ def bn_feature_net_3D(receptive_field=61,
                     kernel_initializer=init, padding='valid',
                     kernel_regularizer=l2(reg))(x[-1]))
     x.append(BatchNormalization(axis=channel_axis)(x[-1]))
-    x.append(Activation('relu')(x[-1]))
+    feature = Activation('relu')(x[-1])
+
+    def __merge_temporal_features(feature, mode='conv', residual=False, n_filters=256,
+                                  n_frames=3, padding=True, temporal_kernel_size=3):
+        if mode == 'conv':
```
willgraf: Maybe do a default value check at the top of the funtion (`if mode is None`).

deepcell/model_zoo/featurenet.py
```diff
@@ -313,7 +314,10 @@ def bn_feature_net_3D(receptive_field=61,
                       padding=False,
                       padding_mode='reflect',
                       multires=False,
-                      include_top=True):
+                      include_top=True,
+                      temporal=None,
```
willgraf: validate this?  If someone uses `"LSTM"` or `"lstm`, should they do the same thing?

deepcell/model_zoo/featurenet.py
```diff
@@ -446,7 +450,39 @@ def bn_feature_net_3D(receptive_field=61,
...
                     kernel_regularizer=l2(reg))(x[-1]))
     x.append(BatchNormalization(axis=channel_axis)(x[-1]))
-    x.append(Activation('relu')(x[-1]))
+    feature = Activation('relu')(x[-1])
+
+    def __merge_temporal_features(feature, mode='conv', residual=False, n_filters=256,
+                                  n_frames=3, padding=True, temporal_kernel_size=3):
+        if mode is None:
+            return feature
+        if mode is 'conv':
```
willgraf: For string comparison, you should keep using `==`, i.e. `if mode == "conv"`.  `is` should be used for object identity. [Here is a stackoverflow post about it](https://stackoverflow.com/questions/6570371/when-to-use-and-when-to-use-is).

I also think it would be a bit cleaner to just `str.lower()` the input, so you do not have to worry about any case mismatches.  With what we have now, `LSTm` would throw an error (in some cases that could be desirable I suppose).

Either way, "conv" mode is still case sensitive.

deepcell/model_zoo/featurenet.py
```diff
@@ -313,7 +314,10 @@ def bn_feature_net_3D(receptive_field=61,
...
                       padding_mode='reflect',
                       multires=False,
-                      include_top=True):
+                      include_top=True,
+                      temporal=None,
+                      residual=False,
+                      temporal_kernel_size=3):
     """Creates a 3D featurenet.
 
     Args:
```
willgraf: You should include your new inputs in the docs as well.

deepcell/model_zoo/featurenet.py
```diff
@@ -470,6 +506,9 @@ def bn_feature_net_3D(receptive_field=61,
 def bn_feature_net_skip_3D(receptive_field=61,
                            input_shape=(5, 256, 256, 1),
                            fgbg_model=None,
+                           temporal=None,
+                           residual=False,
+                           temporal_kernel_size=3,
```
willgraf: Add these to the docstring.

circuit1:
> Almost there! A bit more work on the input validation and mode comparison. Also, please add the new inputs to the docstring.
> 
> Merry Christmas!

Okay, I changed the input for the `bn_feature_net_skip_3D` so the temporal information is encapsulated in the `**kwargs`. I fixed everything else you mentioned. 
[bitmoji](https://render.bitstrips.com/v2/cpanel/a513ff03-2135-43b2-91e1-1267cd4fd097-1e9ea2a3-3f5e-4454-a77a-e75e5e8f42ce-v1.png?transparent=1&palette=1&width=246)

deepcell/model_zoo/featurenet.py
```diff
@@ -446,7 +453,39 @@ def bn_feature_net_3D(receptive_field=61,
...
                     kernel_regularizer=l2(reg))(x[-1]))
     x.append(BatchNormalization(axis=channel_axis)(x[-1]))
-    x.append(Activation('relu')(x[-1]))
+    feature = Activation('relu')(x[-1])
+
+    def __merge_temporal_features(feature, mode='conv', residual=False, n_filters=256,
+                                  n_frames=3, padding=True, temporal_kernel_size=3):
+        if mode is None:
+            return feature
+        if mode.lower() == 'conv':
```
willgraf: I suggest casting mode as a string. If a user unwisely uses `1` as the mode, then we will get an `AttributeError`.  We can fix this by simply casting in the begginning, with `mode = str(mode).lower()`.  Then we can just use simple comparisons, `if mode == 'conv':`

willgraf: Also, when you update this, please update the PR from master.

circuit1: Done!

willgraf: Looks like there is some small linting issue:

```
deepcell-tf/deepcell/model_zoo/featurenet.py:462:1: W293 blank line contains whitespace
```

circuit1: Oops, fixed and ran tests -- should work now?

willgraf:
Great work Sunny!

PR #286, opened 2019-12-23
Speeding up the pixel transform

ngreenwald:
The current implementation of the pixel transform first identifies all unique label IDs in all stacks of training images, then loops over each ID, then over each crop. 

The current implementation of the watershed transform instead first loops over each crop, then identifies the unique cell IDs within each crop, and then loops over each unique label ID.

The latter results in significant speed increases, as only the label IDs within each crop are iterated over. 

This PR refactors the pixel transform in the same style as watershed. 

Obviously a bug here would have far-reaching consequences, so please feel free to go over this with a fine-toothed comb. Happy to make any changes!

### DIFF
--- a/deepcell/image_generators/__init__.py
+++ b/deepcell/image_generators/__init__.py
@@ -96,7 +96,7 @@ def _transform_masks(y, transform, data_format=None, **kwargs):
         edge_class_shape = 4 if separate_edge_classes else 3
 
         if data_format == 'channels_first':
-            y_transform = np.zeros(tuple([y.shape[0]] + list(y.shape[2:]) + [edge_class_shape]))
+            y_transform = np.zeros(tuple([y.shape[0]] + [edge_class_shape] + list(y.shape[2:])))
         else:
             y_transform = np.zeros(tuple(list(y.shape[0:-1]) + [edge_class_shape]))
 
@@ -108,9 +108,6 @@ def _transform_masks(y, transform, data_format=None, **kwargs):
 
             y_transform[batch] = transform_utils.pixelwise_transform(mask, dilation_radius, data_format=data_format,
                                                                      separate_edge_classes=separate_edge_classes)
-
-            if data_format == 'channels_first':
-                y_transform = np.rollaxis(y_transform, y.ndim - 1, 1)
 
     elif transform == 'watershed':
         distance_bins = kwargs.pop('distance_bins', 4)
--- a/deepcell/image_generators/__init__.py
+++ b/deepcell/image_generators/__init__.py
@@ -106,8 +106,9 @@ def _transform_masks(y, transform, data_format=None, **kwargs):
             else:
                 mask = y[batch, ..., 0]
 
-            y_transform[batch] = transform_utils.pixelwise_transform(mask, dilation_radius, data_format=data_format,
-                                                                     separate_edge_classes=separate_edge_classes)
+            y_transform[batch] = transform_utils.pixelwise_transform(
+                mask, dilation_radius, data_format=data_format,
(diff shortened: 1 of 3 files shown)


willgraf:
I haven't had time to review the changes, but it looks like the refactor is not passing the `test_pixelwise_transform_3d` or the `test_pixelwise_transform_3d` tests in `deepcell/utils/transform_utils_test.py`.

You can test this locally using `pytest --cov=deepcell/utils --pep8`.  Make sure you install the `requirements-test.txt` dependencies first.

deepcell/utils/transform_utils.py
```diff
@@ -55,86 +55,70 @@ def pixelwise_transform(mask, dilation_radius=None, data_format=None,
...
         all_stacks = [
-            edges,
-            interiors,
+            edge,
+            interior,
             background
         ]
 
-        return np.stack(all_stacks, axis=channel_axis)
+        return np.stack(all_stacks, axis=-1)
```
willgraf: We still pass in `data_format`, so why are we hardcoding `-1` here?

deepcell/utils/transform_utils.py
```diff
@@ -55,86 +55,70 @@ def pixelwise_transform(mask, dilation_radius=None, data_format=None,
...
-        interior_edges,
-        interiors,
+        background_edge,
+        interior_edge,
+        interior,
         background
     ]
 
-    return np.stack(all_stacks, axis=channel_axis)
+    return np.stack(all_stacks, axis=-1)
```
willgraf: Again, why not leave it as `channel_axis`?

deepcell/image_generators/__init__.py
```diff
@@ -92,10 +92,25 @@ def _transform_masks(y, transform, data_format=None, **kwargs):
...
+        for batch in range(y_transform.shape[0]):
+            if data_format == 'channels_first':
+                mask = y[batch, 0, ...]
+            else:
+                mask = y[batch, ..., 0]
+
+            y_transform[batch] = transform_utils.pixelwise_transform(mask, dilation_radius, data_format=data_format,
+                                                                     separate_edge_classes=separate_edge_classes)
+
+            if data_format == 'channels_first':
```
willgraf: This `rollaxis` is probably unnecessary if we generalize the transform with `data_format`.

deepcell/utils/transform_utils.py
ngreenwald: I removed the channel_axis argument to mirror the logic of the watershed transform, where it's evaluated within the image_generators __init__.py but then not passed into the function: https://github.com/vanvalenlab/deepcell-tf/blob/b78caac6263d46ef1d1570110cb647df0a5a03ed/deepcell/image_generators/__init__.py#L115

If you'd rather not make that same change here happy to revert back

deepcell/utils/transform_utils_test.py
```diff
@@ -105,34 +105,40 @@ def test_pixelwise_transform_3d(self):
             batch_count = maskstack.shape[0] // frames
```
ngreenwald: I wasn't sure why the pixel_transform test went with this approach, compared to the watershed_transform test, so I just looped through the stack for now

ngreenwald:
The tests were failing because of the data format. I changed the tests to also loop through each test image, rather than passing a test image stack.

deepcell/utils/transform_utils_test.py
willgraf: I also do not know why this approach was used, I think what you have done is fine.

willgraf:
Hey this looks great! Thanks for making the changes.  I am going to update from master and merge it in when ready.

Great work!

PR #289, opened 2020-01-01
Manual model savepoints

ngreenwald:
Currently, model weights are automatically saved when the val loss deceases, overwriting previous weight files. I've found that sometimes a decrease in val loss doesn't actually correspond with a better model. I've added an optional flag that species an interval, in epochs, at which the model will be saved.


willgraf:
This is a good idea!

[It looks like TensorFlow has updated the key to `save_freq`](https://github.com/tensorflow/tensorflow/blob/590d6eef7e91a6a7392c8ffffb7b58f2e0c8bc6b/tensorflow/python/keras/callbacks.py#L819). I think we should mimic their naming conventions too (meaning let's change from `manual_save` to `save_freq`) in order to help TF users better use the API.

I think we can also try to pass this in through `kwargs` to prevent having so many inputs to the `train_*` functions (it already has too many).

deepcell/training.py
```diff
@@ -274,7 +278,7 @@ def train_model_conv(model,
         todays_date = datetime.datetime.now().strftime('%Y-%m-%d')
         data_name = os.path.splitext(os.path.basename(dataset))[0]
         model_name = '{}_{}_{}'.format(todays_date, data_name, expt)
-    model_path = os.path.join(model_dir, '{}.h5'.format(model_name))
+    model_path = os.path.join(model_dir, model_name)
```
willgraf: you can also format the model name with the epoch to prevent overwriting model files.

deepcell/utils/train_utils.py
```diff
@@ -54,16 +54,26 @@ def get_callbacks(model_path,
...
     Returns:
         list: a list of callbacks to be passed to model.fit()
     """
-    cbs = [
-        callbacks.ModelCheckpoint(
-            model_path, monitor=monitor,
-            save_best_only=True, verbose=verbose,
-            save_weights_only=save_weights_only),
-    ]
+    if manual_save is None:
```
willgraf: Instead, let's pop the value out of kwargs and default it to the default value in the model checkpoint just to reduce duplicated code.

deepcell/utils/train_utils.py
```diff
@@ -54,16 +54,26 @@ def get_callbacks(model_path,
...
+        cbs = [
+            callbacks.ModelCheckpoint(
+                model_path + ".h5", monitor=monitor,
+                save_best_only=True, verbose=verbose,
+                save_weights_only=save_weights_only),
+        ]
+    else:
+        cbs = [
+            callbacks.ModelCheckpoint(
+                model_path + "_epoch_{epoch:02d}.h5", period=manual_save,
```
willgraf: maybe we can generalize the model name with epoch somehow?

ngreenwald:
> [It looks like TensorFlow has updated the key to `save_freq`](https://github.com/tensorflow/tensorflow/blob/590d6eef7e91a6a7392c8ffffb7b58f2e0c8bc6b/tensorflow/python/keras/callbacks.py#L819). I think we should mimic their naming conventions too (meaning let's change from `manual_save` to `save_freq`) in order to help TF users better use the API.

Yes, I noticed that period is deprecated, and they're suggesting we use save_freq instead. However, save_freq is in units of training examples. The options are either "epoch", in which case every epoch is saved, or an int, which specifies the number of seen examples prior to saving. This means that, for example, in order to request 3 epochs, you would actually need to pass in 3 * number of training examples per epoch.

deepcell/utils/train_utils.py
ngreenwald: Okay @willgraf, so pass manual_save (or save_freq, see my comment above) as part of kwargs. Then get that value from kwargs, rather than defaulting it to None in the training call. We'll still need to duplicate the cbs initialization, right? Given that monitor, save_best_only, and save_freq will be different? Or do you have an idea of how we could set default values for those earlier on that would just get popped out?

deepcell/utils/train_utils.py
ngreenwald: I was wondering about that. We could save the epoch as part of the file name every time the model saves, but that could start taking up a lot of space

ngreenwald:
Hi Will, just wanted to bump this back up. Let me know what you want to do about the save_freq change when you have a chance.

willgraf:
I'm having hesitations merging this in, as I would really just prefer to deprecate the `training.py` scripts. They have too many inputs and really obfuscate what is going on behind the scenes.

ngreenwald:
For sure! In that case I'll just merge it into my own working branch, and once the training.py scripts get refactored we can add this functionality in somewhere to master.

PR #295, opened 2020-01-10
fix pixel transform bug

### LINKED ISSUE #294: pixel transform is broken
Looks like I managed to slip a bug past Will's carefully laid tests. Investigating now.

ngreenwald:
When I modified the pixel transform for #286, I accidentally removed two layers of the for loop. I should have just removed the first. This resulted in only the last cell in each crop being transformed, rather than every cell in the crop. Closes #294

### DIFF
--- a/deepcell/utils/transform_utils_test.py
+++ b/deepcell/utils/transform_utils_test.py
@@ -69,6 +69,7 @@ class TransformUtilsTest(test.TestCase):
 
                 self.assertEqual(pw_img.shape[-1], 3)
                 self.assertEqual(pw_img_dil.shape[-1], 3)
+                assert(np.all(np.equal(pw_img[..., 0] + pw_img[..., 1], img > 0)))
                 self.assertGreater(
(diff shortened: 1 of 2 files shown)


deepcell/utils/transform_utils_test.py
```diff
@@ -86,6 +87,7 @@ def test_pixelwise_transform_2d(self):
 
                 self.assertEqual(pw_img.shape[-1], 4)
                 self.assertEqual(pw_img_dil.shape[-1], 4)
+                assert(np.all(np.equal(pw_img[..., 0] + pw_img[..., 1] + pw_img[..., 2], img > 0)))
```
ngreenwald: I added in an assert statement that makes sure that all pixels which were within a labeled cell are part of interior or border class. I wasn't sure how to do this with self.assert, since we want to do elementwise assertion of the entire array

willgraf: No worries, there's nothing particularly special about `self.assert`.

PR #333, opened 2020-03-27
Change panoptic upsampling method to reduce memory and speed consumption

msschwartz21:
## What
* Add an option to `deepcell.model_zoo.fpn.create_pyramid_levels` that enables choosing between `UpsampleLike` and `Upsampling2D`
* Use `Upsampling2D` in panoptic models instead of `UpsampleLike`
* In `deepcell.model_zoo.panoptic.semantic_upsample` use only the final layer of the pyramid network to blow up to final image size instead of adding together all layers in the pyramid.

## Why
* These changes to upsampling procedures reduces the overall memory footprint of the model and speeds up computation time. The lower memory profile means that this model will be more compatible with hosting multiple models through tf-serving.

### DIFF
--- a/deepcell/model_zoo/panopticnet.py
+++ b/deepcell/model_zoo/panopticnet.py
@@ -293,7 +293,9 @@ def PanopticNet(backbone,
     backbone_dict_reduced = {k: backbone_dict[k] for k in backbone_dict
                              if k in backbone_levels}
 
-    pyramid_dict = create_pyramid_features(backbone_dict_reduced, ndim=2)
+    pyramid_dict = create_pyramid_features(backbone_dict_reduced,
+                                           upsample_type='upsampling2d',
+                                           ndim=2)
 
     semantic_levels = [int(re.findall(r'\d+', k)[0]) for k in pyramid_dict]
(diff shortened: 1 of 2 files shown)


vanvalen:
You should add an option for upsampling3D - the code already does it, but you only allow the upsample_mode to be either upsamplelike or upsampling2D

deepcell/model_zoo/panopticnet.py
```diff
@@ -196,31 +198,52 @@ def __create_semantic_head(pyramid_dict,
...
-        level = int(re.findall(r'\d+', N)[0])
+    # for N, P in zip(pyramid_names, pyramid_features):
+
+    #     # Get level and determine how much to upsample
+    #     level = int(re.findall(r'\d+', N)[0])
+    #     n_upsample = level - target_level
+
+    #     # Use semantic upsample to get semantic map
+    #     semantic_features.append(semantic_upsample_prototype(P, n_upsample, ndim=ndim))
+    #     semantic_names.append('Q{}'.format(level))
```
vanvalen: This commented out section should be deleted

PR #336, opened 2020-03-27
Better input validation for merge temporal features

### LINKED ISSUE #335: __merge_temporal_features should handle bad mode values.
https://github.com/vanvalenlab/deepcell-tf/blob/8040a20bac8cf6eec471e8e5c2ba557f5bbd1c59/deepcell/model_zoo/panopticnet.py#L65-L72

The above should have an else block, or else if this is called with a bad value, then it will crash because `x` was not initialized.

---

https://github.com/vanvalenlab/deepcell-tf/blob/8040a20bac8cf6eec471e8e5c2ba557f5bbd1c59/deepcell/model_zoo/panopticnet.py#L368-L371

Additionally, we should be better about input validation here. Cast `mode` to a lowercase string in `__merge_temporal_features` or here where we call `__merge_temporal_features`. Finally, a set would be a better comparison `feature in {'abc', 'def', '123}`, O(1) vs O(n).

msschwartz21:
## What
* Add better input validation for `_merge_temporal_features` function for panoptic networks. Cast to lowercase and check the input before running the function itself or using it to build a model

## Why
* Prevents uncontrolled crashes due to bad user input

Closes #335

### DIFF
--- a/deepcell/model_zoo/panopticnet.py
+++ b/deepcell/model_zoo/panopticnet.py
@@ -48,7 +48,27 @@ from deepcell.utils.misc_utils import get_sorted_keys
 
 
 def __merge_temporal_features(feature, mode='conv', feature_size=256, frames_per_batch=1):
-    if mode == 'conv':
+    """
+
+    Args:
+        feature: Input layer
+        mode (str, optional): Choose from {'conv','lstm','gru'}. Defaults to 'conv'.
+        feature_size (int, optional): Defaults to 256.
+        frames_per_batch (int, optional): Defaults to 1.
+
+    Raises:
+        ValueError: Mode not 'conv', 'lstm' or 'gru'
+
+    Returns:
+        Same as feature: Output of layer following merge of temporal features
+    """
+
+    acceptable_modes = {'conv', 'lstm', 'gru'}
+    if mode.lower() not in acceptable_modes:
+        raise ValueError('Mode {} not supported. Please choose from {}.'.format(
+            mode.lower(), str(acceptable_modes)))
+
+    if mode.lower() == 'conv':
         x = Conv3D(feature_size,
                    (frames_per_batch, 3, 3),
                    strides=(1, 1, 1),
@@ -56,13 +76,13 @@ def __merge_temporal_features(feature, mode='conv', feature_size=256, frames_per
                    )(feature)
         x = BatchNormalization(axis=-1)(x)
         x = Activation('relu')(x)
-    elif mode == 'lstm':
+    elif mode.lower() == 'lstm':
         x = ConvLSTM2D(feature_size,
                        (3, 3),
                        padding='same',
                        activation='relu',
                        return_sequences=True)(feature)
-    elif mode == 'gru':
+    elif mode.lower() == 'gru':
         x = ConvGRU2D(feature_size,
                       (3, 3),
                       padding='same',
@@ -299,10 +319,18 @@ def PanopticNet(backbone,
             backbone.  3 is the default for all current backbones.
         kwargs (dict): Other standard inputs for retinanet_mask.
 
+    Raises:
+        ValueError: temporal_ode not 'conv', 'lstm' or 'gru'
+
     Returns:
         tensorflow.keras.Model: Panoptic model with a backbone.
     """
     channel_axis = 1 if K.image_data_format() == 'channels_first' else -1
+
+    acceptable_modes = {'conv', 'lstm', 'gru'}
+    if temporal_mode.lower() not in acceptable_modes:
+        raise ValueError('Mode {} not supported. Please choose from {}.'.format(
+            temporal_mode.lower(), str(acceptable_modes)))
 
     if inputs is None:
         if frames_per_batch > 1:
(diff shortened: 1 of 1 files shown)


deepcell/model_zoo/panopticnet.py
```diff
@@ -48,21 +48,41 @@
 
 
 def __merge_temporal_features(feature, mode='conv', feature_size=256, frames_per_batch=1):
-    if mode == 'conv':
+    """
+
```
willgraf: remove whitepsace

deepcell/model_zoo/panopticnet.py
```diff
@@ -48,21 +48,41 @@
...
+    Raises:
+        ValueError: Mode not 'conv', 'lstm' or 'gru'
+
+    Returns:
+        Same as feature: Output of layer following merge of temporal features
+    """
+
+    acceptable_modes = {'conv', 'lstm', 'gru'}
+    if mode.lower() not in acceptable_modes:
```
willgraf: what if mode is not a string? maybe something like:
`if mode and str(mode).lower() not in....`

Same assumption down below in the ValueError

deepcell/model_zoo/panopticnet.py
```diff
@@ -48,21 +48,41 @@
...
         x = Activation('relu')(x)
-    elif mode == 'lstm':
+    elif mode.lower() == 'lstm':
         x = ConvLSTM2D(feature_size,
                        (3, 3),
                        padding='same',
                        activation='relu',
                        return_sequences=True)(feature)
-    elif mode == 'gru':
+    elif mode.lower() == 'gru':
```
willgraf: We still need an else block or else `x` is not defined.

deepcell/model_zoo/panopticnet.py
```diff
@@ -48,21 +48,41 @@
...
+    Returns:
+        Same as feature: Output of layer following merge of temporal features
+    """
+
+    acceptable_modes = {'conv', 'lstm', 'gru'}
+    if mode.lower() not in acceptable_modes:
+        raise ValueError('Mode {} not supported. Please choose from {}.'.format(
+            mode.lower(), str(acceptable_modes)))
+
+    if mode.lower() == 'conv':
```
willgraf: maybe overwrite mode above so you don't need to keep calling `.lower()`

deepcell/model_zoo/panopticnet.py
```diff
@@ -299,11 +319,19 @@ def PanopticNet(backbone,
             backbone.  3 is the default for all current backbones.
         kwargs (dict): Other standard inputs for retinanet_mask.
 
+    Raises:
+        ValueError: temporal_ode not 'conv', 'lstm' or 'gru'
```
willgraf: typo

deepcell/model_zoo/panopticnet.py
```diff
@@ -299,11 +319,19 @@ def PanopticNet(backbone,
...
     Returns:
         tensorflow.keras.Model: Panoptic model with a backbone.
     """
     channel_axis = 1 if K.image_data_format() == 'channels_first' else -1
 
+    acceptable_modes = {'conv', 'lstm', 'gru'}
+    if temporal_mode.lower() not in acceptable_modes:
+        raise ValueError('Mode {} not supported. Please choose from {}.'.format(
+            temporal_mode.lower(), str(acceptable_modes)))
+
```
willgraf: more type assumptions that may fail.

deepcell/model_zoo/panopticnet.py
```diff
@@ -366,11 +394,10 @@ def PanopticNet(backbone,
...
 
     if frames_per_batch > 1:
-        if temporal_mode in ['conv', 'lstm', 'gru']:
-            temporal_features = [__merge_temporal_features(
-                feature, mode=temporal_mode) for feature in features]
-            for f, k in zip(temporal_features, pyramid_dict.keys()):
-                pyramid_dict[k] = f
+        temporal_features = [__merge_temporal_features(
+            feature, mode=temporal_mode) for feature in features]
+        for f, k in zip(temporal_features, pyramid_dict.keys()):
```
willgraf: Does this give us the expected answer if `temporal_mode` is `None` (the default value)?

deepcell/model_zoo/panopticnet.py
msschwartz21: Can I eliminate this concern by starting with `mode = str(mode).lower()`?

deepcell/model_zoo/panopticnet.py
msschwartz21: I'm not totally sure how `x` should be defined in the else case. Would you want it to return `feature` without any modification? I thought that by putting in `raise ValueError` that we would avoid getting to the condition where else is triggered.

deepcell/model_zoo/panopticnet.py
willgraf: yeah maybe put it in an if block, so if `mode` is `None`, it stays `None`

Maybe `mode = str(mode).lower() if mode else mode`

willgraf: But I think we should handle if `mode is None` because it does that by default currently.

deepcell/model_zoo/panopticnet.py
willgraf: We would avoid that for now, but it is just an opening for a `NullPointerError`. This is a coding best-practice issue.  If `mode is None`, maybe we should just return the unmodified feature?

deepcell/model_zoo/panopticnet.py
msschwartz21: Is the assumption here that `temporal_mode` is a string?

willgraf: Yes

PR #341, opened 2020-04-02
Update applications with a new SegmentationApplication base class

msschwartz21:
## What
* Model objects for nuclear and cytoplasmic segmentation are replaced with application objects

## Why
* Provides more comprehensive features including a predict function that handles tiling along with pre/post processing functions for the data


msschwartz21:
The build is currently failing because it is dependent on the master branch of deepcell-toolbox.

deepcell/applications/cytoplasm_segmentation.py
```diff
@@ -0,0 +1,96 @@
...
+            'batch_size': 2,
+            'lr': 1e-5,
+            'lr_decay': 0.95,
+            'training_seed': 0,
+            'n_epochs': 8,
+            'training_steps_per_epoch': 7899 // 2,
+            'validation_steps_per_epoch': 1973 // 2
+        }
+
+        super(CytoplasmSegmentationApplication, self).__init__(self.model, **dict(
```
willgraf, line 87: Don't unzip a dict, just use key-value pairs.

deepcell/applications/nuclear_segmentation.py
```diff
@@ -23,66 +23,75 @@
...
+            'lr': 1e-4,
+            'lr_decay': 0.95,
+            'training_seed': 0,
+            'n_epochs': 8,
+            'training_steps_per_epoch': 82800 // 16,
+            'validation_steps_per_epoch': 20760 // 16
+        }
 
-    return model
+        super(NuclearSegmentationApplication, self).__init__(self.model, **dict(
```
willgraf, line 88: Here too, no need to create a dict and then unzip it.

deepcell/applications/segmentation_application.py
```diff
@@ -0,0 +1,140 @@
...
+            new_shape = (int(image.shape[1] / scale_factor),
+                         int(image.shape[2] / scale_factor))
+            image = resize(image, new_shape, data_format='channels_last')
+        else:
+            original_shape = None
+
+        # Preprocess image
+        if self.preprocessing_fn is not None:
+            image = self.preprocessing_fn(image, **preprocess_kwargs)
+            print('preprocessing')
```
willgraf, line 115: Maybe also check if it's "callable", in case users don't pass a function in.

deepcell/applications/segmentation_application.py
```diff
@@ -0,0 +1,140 @@
...
+
+        Returns:
+            np.array: Labeled image, if debug is False.
+                If debug is True, returns (image, tiles, output_tiles, output_images, label_image)
+        """
+
+        # Check input size of image
+        if len(image.shape) != 4:
+            raise ValueError('Input data must have 4 dimensions, [batch, x, y, channel'
+                             'Input data only has {} dimensions'.format(str(len(image.shape))))
```
willgraf, line 100: Instead of `4`, maybe we could use a more general solution, like `required_rank`.

Or get it out of the `model.input_shape`?

deepcell/applications/segmentation_application.py
```diff
@@ -0,0 +1,140 @@
...
+            label_image = output_images[0]
+
+        # Resize label_image back to original resolution if necessary
+        if original_shape is not None:
+            label_image = resize(label_image, original_shape[1:3], data_format='channels_last')
+
+        if debug:
+            return image, tiles, output_tiles, output_images, label_image
+        else:
+            return label_image
```
willgraf, line 140: ???

deepcell/applications/nuclear_segmentation.py
```diff
@@ -23,66 +23,75 @@
...
+                                 location=True,
+                                 include_top=True)
 
-    if use_pretrained_weights:
-        if backbone == 'resnet50':
-            # '/data/models/resnet50_panoptic_train-val_DVV_V4.h5'
+        if use_pretrained_weights:
             weights_path = get_file(
-                'resnet50_panoptic_nuclear_segmentation.h5',
+                WEIGHTS_PATH.split('/')[-1],
```
willgraf, line 63: maybe use `os.path.basename` to make it a bit more general.

deepcell/applications/segmentation_application.py
msschwartz21: Added this test and raising error if it doesn't pass.

deepcell/applications/segmentation_application.py
msschwartz21: I added this option so that we can extract the intermediate processing steps in case we're troubleshooting. I end up using it to validate that the preprocessing function is working.

deepcell/applications/segmentation_application.py
msschwartz21: Changed to compare to `self.required_rank` defined as `len(self.model_input_shape) + 1`

deepcell/applications/application.py
```diff
@@ -0,0 +1,164 @@
...
+        # Test that pre and post processing functions are callable
+        if (self.preprocessing_fn is not None) and (callable(self.preprocessing_fn) is False):
+            raise ValueError('Preprocessing_fn must be a callable function.')
+        if (self.postprocessing_fn is not None) and (callable(self.postprocessing_fn) is False):
+            raise ValueError('Postprocessing_fn must be a callable function.')
+
+    def predict(self, x):
+        raise NotImplementedError
+
+    def _predict_segmentation(self,
```
willgraf, line 89: Will all `Application` children have a segmentation prediction? If not, maybe this logic belongs in a child class.  You could have another class `SegmentationApplication` which inherits from `Application`.

deepcell/applications/application.py
```diff
@@ -0,0 +1,164 @@
...
+            new_shape = (int(image.shape[1] / scale_factor),
+                         int(image.shape[2] / scale_factor))
+            image = resize(image, new_shape, data_format='channels_last')
+        else:
+            original_shape = None
+
+        # Preprocess image
+        if self.preprocessing_fn is not None:
+            image = self.preprocessing_fn(image, **preprocess_kwargs)
+            print('preprocessing')
```
willgraf, line 139: I would prefer not to have these print statements. If necessary, use a logger.

deepcell/applications/cytoplasm_segmentation.py
```diff
@@ -0,0 +1,97 @@
...
+                 use_pretrained_weights=True,
+                 model_image_shape=(128, 128, 1)):
+
+        self.model = PanopticNet('resnet50',
+                                 input_shape=model_image_shape,
+                                 norm_method='whole_image',
+                                 num_semantic_heads=3,
+                                 num_semantic_classes=[1, 1, 2],
+                                 location=True,
+                                 include_top=True)
```
willgraf, line 60: You don't need to assign this to self, as it gets passed to the superclass init, which also assigns to self.

deepcell/applications/nuclear_segmentation.py
```diff
@@ -23,66 +23,76 @@
...
+                 use_pretrained_weights=True,
+                 model_image_shape=(128, 128, 1)):
+
+        self.model = PanopticNet('resnet50',
+                                 input_shape=model_image_shape,
+                                 norm_method='whole_image',
+                                 num_semantic_heads=3,
+                                 num_semantic_classes=[1, 1, 2],
+                                 location=True,
+                                 include_top=True)
```
willgraf, line 61: See other comment.

deepcell/applications/nuclear_segmentation.py
```diff
@@ -23,66 +23,76 @@
...
+        super(NuclearSegmentation, self).__init__(self.model,
+                                                  model_image_shape=model_image_shape,
+                                                  model_mpp=0.65,
+                                                  preprocessing_fn=None,
+                                                  postprocessing_fn=deep_watershed,
+                                                  dataset_metadata=dataset_metadata,
+                                                  model_metadata=model_metadata)
 
-    return model
+        self.predict = self._predict_segmentation
```
willgraf, line 98: I think it would be clearer to redefine `predict` than to assign it in the init.

deepcell/applications/cytoplasm_segmentation.py
```diff
@@ -0,0 +1,97 @@
...
+
+        super(CytoplasmSegmentation, self).__init__(self.model,
+                                                    model_image_shape=model_image_shape,
+                                                    model_mpp=0.65,
+                                                    preprocessing_fn=None,
+                                                    postprocessing_fn=deep_watershed,
+                                                    dataset_metadata=dataset_metadata,
+                                                    model_metadata=model_metadata)
+
+        self.predict = self._predict_segmentation
```
willgraf, line 97: I think it would be clearer to redefine `predict` than to assign it in the init.

deepcell/applications/application.py
msschwartz21: My bad, left over from debugging.

deepcell/applications/application.py
msschwartz21: Yes all application objects should have some form of a predict function. I was reluctant to create another intermediate class because then we would end up with a third layer of inheritance that I don't think adds any clarity.

deepcell/applications/application.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Postprocess predictions to create label image
+        if self.postprocessing_fn is not None:
+            label_image = self.postprocessing_fn(output_images, **postprocess_kwargs)
+        else:
+            label_image = output_images[0]
+
+        # Resize label_image back to original resolution if necessary
+        if original_shape is not None:
+            label_image = resize(label_image, original_shape[1:3], data_format='channels_last')
```
willgraf, line 159: The numbers `[1:3]` make me a little nervous. Are these values related to `required_rank`?

deepcell/applications/application.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # Untile images
+        output_images = [untile_image(o, tiles_info, model_input_shape=self.model_image_shape,
+                                      dtype=o.dtype) for o in output_tiles]
+
+        # Postprocess predictions to create label image
+        if self.postprocessing_fn is not None:
+            label_image = self.postprocessing_fn(output_images, **postprocess_kwargs)
+        else:
+            label_image = output_images[0]
```
willgraf, line 155: why only the first value?

deepcell/applications/nuclear_segmentation.py
```diff
@@ -23,66 +23,77 @@
...
 from __future__ import print_function
 
+import os
+
 from tensorflow.python.keras.utils.data_utils import get_file
 
+from deepcell_toolbox.deep_watershed import deep_watershed
+
 from deepcell.utils.retinanet_anchor_utils import generate_anchor_params
 from deepcell import model_zoo
```
willgraf, line 39: remove unused imports.

deepcell/applications/application.py
msschwartz21: When we untile we use list comprehension, so the output always ends up in a list. If there is only a single image as output then line 155 essentially takes it out of the list. 

The inherent assumption is that after running postprocessing there should only be a single image. We rely on that assumption during the resize test. I'm not sure if it's appropriate to generalize this further. Thoughts?

deepcell/applications/application.py
msschwartz21: Yes somewhat. Starting from 1 is a result of dropping the batch dimension. I could change it to `original_shape[1:]` which would generalize to rank larger than 4.

deepcell/applications/nuclear_segmentation_test.py
```diff
@@ -23,40 +23,39 @@
...
 
+        self.app = NuclearSegmentation(use_pretrained_weights=False)
 
-class TestNuclearSegmentationModel(test.TestCase):
+    def test_nuclear_app(self):
 
-    def test_nuclear_segmentation_model(self):
+        # Check shape parameters
+        shape = self.app.model.output_shape
+        print(shape)
```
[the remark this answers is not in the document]
msschwartz21: Also removed from cytoplasm testing

deepcell/applications/application.py
msschwartz21: This is no longer an issue after creating data processing utility methods.

PR #342, opened 2020-04-03
Expand metrics package functionality

### LINKED ISSUE #293: Metrics package should retain information about cell identity
Currently, compiling metrics removes all the information about which cells are segmented correctly and which have errors. We should correct this and have some tool to visualize where errors occur.

### LINKED ISSUE #314: Metrics package should identify which cells have segmentation errors
The metrics package only tells you how many of each error type occurs. It should identify each cell that has a segmentation error so the mistakes can be visualized.

Return id's of cells that have segmentation errors

None

**Additional context**
None

### LINKED ISSUE #339: IOU threshold in metrics package results in merges being misclassified as misses
The current metrics package uses IOU as a threshold for determining when a predicted cell and a true cell should be linked together. However, if one cell is much smaller than the other cell, even if the the small one is completely contained within the larger cell, the IOU will be quite small. 

For example, here's a test function Morgan wrote that generates 3 ground truth cells:
image

It then generates an erroneous prediction of two of the cells merged together:
image

However, because the aqua cell in the top left of the true labels is quite small, the IOU between it and the yellow cell in the predicted cell is only 0.06. Thus, it gets discarded as a linked cell for the large cell. 

This means that during assignment, the large yellow predicted cell is only linked with the large yellow true cell. Therefore, the yellow-yellow link is counted as a correct prediction, the green-green link is counted a correct prediction, and the aqua cell is counted as "missed" in the predicted labels.

Morgan and I have been discussing possible ways to get around this. 

One way to address this would be to lower the threshold for determining when two cells are linked. However, this will result in false positives when two large cells happen to overlap each other. 

Another way to address this would be to use a combination of IOU and something else to construct the matrix. For example, you could compute the % of cell1 that is within cell2 for all cells with a non-zero IOU, and then take the max of that value and IOU. In this case, if 100% (or even 40%) of a small cell was contained within a large cell, instead of an IOU of .06 it would get bumped to 0.4, which would be well clear of the threshold. 

If you all have other ideas for ways we could address this would love to hear it!

ngreenwald:
## What
* metrics now tests for correctness of assigned error types rather than just lack of run-time errors
* metrics now returns the label ID of cells assigned to each category
* metrics now plots the segmentation mask of all cells in the crop, color-coded by the error type
* metrics now automatically adds small cells which were removed from a multi-cell merge event
* metrics now optionally will ensure that any cell which has a small cell merged into it is not classified as a correct prediction with an unrelated missed detection, but instead as a merge event between those cells. 

## Why
* returning error labels is helpful for visualization purposes
* checking these labels identified a couple bugs in the assignment script
* the optional flag to ensure that merge events are not misclassified as accurate predictions with missed events can be useful when cells have a large range of sizes

closes #293, closes #314, closes #339


deepcell/metrics.py
```diff
@@ -157,14 +159,14 @@ class ObjectAccuracy(object):  # pylint: disable=useless-object-inheritance
...
-    # TODO: Implement recording of object indices for each error group
     def __init__(self,
                  y_true,
                  y_pred,
                  cutoff1=0.4,
                  cutoff2=0.1,
                  test=False,
-                 seg=False):
+                 seg=False,
+                 force_event_links=False):
```
msschwartz21, line 169: Update the docstring to include documentation of this parameter. I think it would also be appropriate in the general description of the function to describe when a user would want to use `force_event_links`.

deepcell/metrics.py
```diff
@@ -213,9 +215,11 @@ def __init__(self,
 
         self.merge_indices = {}
         self.merge_indices['y_true'] = []
+        self.merge_indices['y_pred'] = []
```
msschwartz21, line 218: I think this is a slightly cleaner way to define these dictionaries.
```suggestion
        self.merge_indices = {
            'y_true': [],
            'y_pred': []
        }
```

deepcell/metrics.py
```diff
@@ -213,9 +215,11 @@ def __init__(self,
...
 
         self.split_indices = {}
         self.split_indices['y_true'] = []
+        self.split_indices['y_pred'] = []
```
msschwartz21, line 222:
```suggestion
        self.split_indices = {
            'y_true': [],
            'y_pred': []
        }
```

deepcell/metrics.py
```diff
@@ -213,9 +215,11 @@ def __init__(self,
...
 
         self.catastrophe_indices = {}
         self.catastrophe_indices['y_true'] = []
```
msschwartz21, line 225:
```suggestion
        self.catastrophe_indices = {
            'y_true': []
        }
```

deepcell/metrics.py
```diff
@@ -738,7 +802,12 @@ def calc_object_stats(self, y_true, y_pred):
             y_true (numpy.array): Labeled ground truth annotations
             y_pred (numpy.array): Labeled prediction mask
         """
+
+        if len(y_true.shape) < 3:
+            raise ValueError("Invalid input dimensions: must be at least 3D tensor")
```
msschwartz21, line 807: Add documentation for this error to the docstring

deepcell/metrics.py
```diff
@@ -982,3 +1054,40 @@ def match_nodes(gt, res):
             iou[frame, iou_gt_idx, iou_res_idx] = intersection.sum() / union.sum()
 
     return iou
+
+
+def plot_errors(y_true, y_pred, error_dict):
+    """Plots the errors identified from linear assignment code
```
msschwartz21, line 1060: Document the inputs for this function

deepcell/metrics.py
```diff
@@ -293,6 +298,45 @@ def get_box_labels(images):
                (intersection.sum() > 0.5 * np.sum(self.y_true == index)):
                 self.seg_thresh[iou_y_true_idx - 1, iou_y_pred_idx - 1] = 1
 
+    def _modify_iou(self, force_event_links):
+        """Modifies the IOU matrix to boost the value for small cells.
```
msschwartz21, line 302: Add input parameter documentation

deepcell/metrics.py
```diff
@@ -157,14 +159,14 @@ class ObjectAccuracy(object):  # pylint: disable=useless-object-inheritance
     Warning:
         Position indicies are not currently collected appropriately
```
msschwartz21, line 160: Remove this warning since you have now fixed the problem.

deepcell/metrics.py
ngreenwald: Got it, let me know if that looks like enough detail or if I should add more

ngreenwald:
Couple open questions. What do you think we should do about predictions not being relabeled when they're supplied? Do we reject non-sequential inputs? Or relabel them ourselves and return the relabeled version?

What is the best way to test the plotting function? Will travis error out if there's nowhere to actually plot something when the code gets tested?

msschwartz21:
Ah got it. Let's just return the relabeled images so that we know the labels are consistent.

.travis.yml
```diff
@@ -12,7 +12,7 @@ python:
   - 3.7
 
 env:
-  - TF_VERSION=1.14.0
+  - TF_VERSION=1.14.0
```
willgraf, line 15: please remove the trailing whitespace.

deepcell/metrics.py
```diff
@@ -339,8 +387,8 @@ def _linear_assignment(self):
         # Identify direct matches as true positives
         correct_index = np.where(self.cm_res[:self.n_true, :self.n_pred] == 1)
         self.correct_detections += len(correct_index[0])
-        self.correct_indices['y_true'].append(correct_index[0])
-        self.correct_indices['y_pred'].append(correct_index[1])
+        self.correct_indices['y_true'] += list(correct_index[0] + 1)
+        self.correct_indices['y_pred'] += list(correct_index[1] + 1)
```
willgraf, line 391: Concatenating lists is slow. Try using `list.extend` (for adding multiple values) or `list.append` (for adding single values) instead.

deepcell/metrics.py
```diff
@@ -544,6 +598,20 @@ def save_to_dataframe(self):
...
+        """Saves the ids of cells in each error category for subsequent visualization
+
+        Returns:
+            error_dict: dictionary containing {category_name: id list} pairs
+        """
+
+        error_dict = {"splits": self.split_indices, "merges": self.merge_indices,
+                      "gains": self.gained_indices, "misses": self.missed_indices,
+                      "catastrophes": self.catastrophe_indices,
+                      "correct": self.correct_indices}
```
willgraf, line 611: can we put each on its own line for readability?

deepcell/metrics.py
ngreenwald: Yeah, that's what morgan had before: 
image

However, it was resulting in an output that was an array, and inside the array a list, and inside the list the values we wanted: [list([1, 2, 4)]. Sometimes we're adding a single value and sometimes multiple.

willgraf: What are the possible value types of `correct_index[0]`? Maybe we can enforce that it is always the same type? (always cast to a list, then call `.extend()`)

We should avoid using `list += list()` as it is very slow and creates duplicate lists.

ngreenwald: Got it. It's always a np.array, so we should always be able to cast to list. Changing now

ngreenwald:
The build passed, for some reason Travis isn't updating. Let me know if there's anything else Will

deepcell/metrics.py
```diff
@@ -293,6 +298,49 @@ def get_box_labels(images):
...
+
+                # optionally, we can also decrease the IOU value of the cell that
+                # swallowed up the small cell so that it doesn't directly match a different cell
+                if force_event_links:
+                    if true_in_pred > 0.5:
+                        fix_idx = np.where(self.iou[:, pred_label - 1] > 1 - self.cutoff1)
+                        self.iou_modified[fix_idx, pred_label - 1] = 1 - self.cutoff1 - 0.01
+                    elif pred_in_true > 0.5:
+                            fix_idx = np.where(self.iou[true_label - 1, :] > 1 - self.cutoff1)
+                            self.iou_modified[true_label - 1, fix_idx] = 1 - self.cutoff1 - 0.01
```
willgraf, line 342: the whitespace looks weird.

deepcell/metrics.py
```diff
@@ -727,7 +797,7 @@ def print_pixel_report(self):
         print('\nConfusion Matrix')
         print(self.cm)
 
-    def calc_object_stats(self, y_true, y_pred):
+    def calc_object_stats(self, y_true, y_pred, return_predictions=False):
```
willgraf, line 800: why don't we just always return the predictions?

ngreenwald: Right, I guess the question is what do you want the default behavior to be. You think it would be useful to always have this return? I was thinking that when you just want to print the metrics summary after training a new model you might not want anything to be returned? But I suppose it's equally easy to assign to _ if you don't want the output? Whatever you think makes more sense.

willgraf: You don't even need to assign it to anything if you don't want the return value.

To me, it just seems like unnecessary complexity. @msschwartz21 do you have any any thoughts on this?

msschwartz21: I would lean towards not returning it at all since someone can always use `self.predictions` if they want access to the data. Would the primary utility be for visualizing mistakes?

msschwartz21:
Just to confirm, are we planning on merging this PR into deepcell-tf and then migrating metrics to the toolbox?

ngreenwald:
Yup, that was the plan

deepcell/metrics.py
ngreenwald: Yes, the reason to return IDs would be for visualization or analysis using the label IDs

willgraf: I'm not sure whether or not any of the return is necessary, I'll leave that to you.

Whether or not we want to return data, we should always do it. let's remove the flag that controls whether or not something is returned.

ngreenwald:
Great! Can one of you merge this in when you have a chance?

PR #344, opened 2020-04-06
Training on multiple labels of the same image

ngreenwald:
## What
Modified SemanticDataLoader to take y and y_2 labels, apply separate set of transforms to y2. 

## Why
Allows network to be trained with different labels. 

First attempt at training a deep watershed network worked as expected! You can see that it's predicting a slightly more narrow, nucleus-focused range for this semantic head:
Screen Shot 2020-04-06 at 3 39 30 PM
compared to this one, which is predicting a slightly broader whole cell:

Screen Shot 2020-04-06 at 3 39 37 PM

The nuclear labels I used are fairly noisy, since they're the result of thresholding the imaging data, creating a mask, and zeroing out the whole cell labels based on that mask. More accurate annotations would definitely help accentuate the difference. 

## Next steps
Given that it can work in theory,  what would be the best way to implement this. Does a separate list of y_transforms for the second set of labels make sense? Would it be better to turn y_transforms into a list of lists, and if there's only one list default to current implementation?

What about for passing the data into the data loader? Should we call the second set of labels y_2? Obviously we'll only want to attempt to get it from dictionary if it's been supplied. 

The deep watershed network as it currently exists is just cells in a jupyter notebook, which makes swapping in an extra set of labels quite easy. Once it gets pulled into a defined model, what would be the best way to add this without introducing a lot of overhead? 

Or is this more of a niche application that's specific for MIBI data, and it doesn't make sense to  introduce this into master?


deepcell/image_generators/semantic.py
```diff
@@ -1,387 +1,403 @@
...
+        save_format (str): Format to use for saving sample images
+            (if save_to_dir is set).
+    """
+    def __init__(self,
+                 train_dict,
+                 image_data_generator,
+                 batch_size=1,
+                 shuffle=False,
+                 transforms=['watershed-cont'],
+                 transforms_y2=['pixelwise'],
```
ngreenwald, line 81: For some reason git is freaking out and said the entire file changed. First change is here

deepcell/image_generators/semantic.py
```diff
@@ -1,387 +1,403 @@
...
+            y_transform = _transform_masks(y, transform,
+                                           data_format=data_format,
+                                           **transform_kwargs)
+            if y_transform.shape[self.channel_axis] > 1:
+                y_transform = np.asarray(y_transform, dtype='int32')
+            elif y_transform.shape[self.channel_axis] == 1:
+                y_transform = np.asarray(y_transform, dtype=K.floatx())
+            self.y_semantic_list.append(y_transform)
+
+        # Add transformed masks for second set of labels
```
ngreenwald, line 132: And here

ngreenwald:
Hey Dave, Will and I fixed the issue with the entire file showing up in the diff. 
@willgraf, maybe you could give some comments as well about how best to integrate the option for multiple labels in as seamless a manner as possible? I put some suggestions at the top, obviously totally open to your feedback

willgraf:
I definitely think that a list of lists is preferable to `transforms_y2`, as this only really solves it for 2 labels, not "multiple labels". I need to think about it a bit more.

ngreenwald:
Okay, I switched the format so that label-specific transforms are specified with lists of lists. If a single list is passed but there are multiple labels, it will generate those same semantic heads for each distinct set of labels. 

I also switched the data format so that when there are multiple labels, 'y' within the train_dict is a dictionary that holds the additional labels. My thinking is that this would allow all of the current functionality you've built that assumes 'X' and 'y' are the only keys in the train_dict to not have to be modified, and it's only once it gets to this stage that we discover if it's a single set of labels or not. 

Maybe @dylanbannon wants to weigh in as well if that will work with new training framework? I took a peek and it seemed like it would. 

Open to other suggestions as well!

ngreenwald:
Hey @vanvalen, let me know what you think. It's much cleaner now. 

Will brought up a good point: are there hidden y_true[..., 0] assumptions hiding somewhere in the code? Anything after the data loader will be fine, since it loops through each label axis and the result is the same shape as before, but if there's anything before it could fail silently.

ngreenwald:
@willgraf @vanvalen This is looking really nice with the new model. Do you want me to add this in as a separate class or include it in the semantic data generator?

willgraf:
I'll let @vanvalen share his thoughts too, but it looks like it's not changing that much really, so I'd be alright with leaving this in the `SemanticGenerator` class.

deepcell/image_generators/semantic.py
```diff
@@ -117,26 +118,32 @@ def __init__(self,
...
+
+        # loop over channels axis of labels in case there are multiple label types
+        for label_num in range(y.shape[self.channel_axis]):
+
+            if self.channel_axis == 1:
+                y_current = y[:, label_num:(label_num + 1), ...]
+            else:
+                y_current = y[..., label_num:(label_num + 1)]
+
+            print('shape of y_current is {}'.format(y_current.shape))
```
willgraf, line 130:
```suggestion
```

No need for the `print`

deepcell/image_generators/image_generators_test.py
```diff
@@ -38,6 +38,7 @@
 from tensorflow.python.platform import test
 
 from deepcell import image_generators
+from deepcell.image_generators import semantic
```
willgraf, line 41: We can remove this and import the `SemanticDataGenerator` directly from `image_generators` instead of importing `semantic`.

PR #345, opened 2020-04-07
Model trainer

dylanbannon:
## What
An integrated framework for training all Deepcell models.

## Why
Right now, training procedures vary by model, which can make training difficult for users new to a specific model. Having a unified framework will help users not to get tripped up by model idiosyncracies while training.

## Current state (will be updated as progress continues)
Scripts for testing the `ModelTrainer` class' ability to 1) train models, and 2) compare models are present in the `deepcell/training` folder.


willgraf:
I recommend having a `model_trainer_test.py` file that tests each of the class' methods.

dylanbannon:
Just a side note: I looked into Tensorflow Extended (https://www.tensorflow.org/tfx) and it is basically exactly what we're looking for here, plus some extra TF-serving-related stuff. In total, it's an ontology of everything that might be involved in the training and deployment of neural nets.

To me, it sounds like it's a great long-term solution for the main problem we're trying to solve with this branch: reproducibility of trained neural nets. However, it would involve adding annotation code to every single object in the deepcell library and, because of that, I really don't feel qualified to be implementing it myself.

So, unless there are other opinions on this matter, I'm going to implement our metadata export solutoin to help us keep track of how models were generated, as planned, and suggest that Tensorflow Extended might be a good way to replace the work I'm doing now, if we have time to implement something comprehensive in the future.

deepcell/training/model_reconstitution_test.py
```diff
@@ -0,0 +1,198 @@
...
+import json
+import hashlib
+import re
+
+import os
+import random
+import numpy as np
+import tensorflow as tf
+
+#import deepcell_toolbox
```
willgraf, line 34: PEP8 will want a space between the `#` and your comment.

deepcell/training/model_reconstitution_test.py
```diff
@@ -0,0 +1,198 @@
...
+        assert self.source_metadata["deepcell_version"] == deepcell.__version__
+
+    def _load_data(self):
+        # take in dataset
+        filename = 'HeLa_S3.npz'
+        test_size = 0.1 # % of data saved as test
+        (self.X_train, self.y_train), (self.X_test, self.y_test) = deepcell.datasets.hela_s3.load_data(
+                                                                   filename,
+                                                                   test_size = test_size,
+                                                                   seed = self.seed_value)
```
willgraf, line 52: It's best to create all class attributes in `__init__`. Without initializing attributes, the class could crash if it calls methods out of order.

deepcell/training/model_reconstitution_test.py
```diff
@@ -0,0 +1,198 @@
...
+        #validation_iterator = deepcell.ImageFullyConvIterator(
+        #                    train_dict = validation_dict,
+        #                    image_data_generator = train_generator)
+
+        ## read in parameters from metadata
+        for generator_type in ("train_generator", "validation_generator"):
+            if self.source_metadata["generators"][generator_type]["parameters"]["seed"]:
+                self.generator_seed = self.source_metadata["generators"][generator_type]["parameters"]["seed"]
+            else:
+                self.generator_seed = 43
```
willgraf, line 85: maybe this should be a default_seed or something configurable instead of just `43`.

deepcell/training/model_reconstitution_test.py
```diff
@@ -0,0 +1,198 @@
...
+                    n_conv_filters=32,
+                    n_dense_filters=128,
+                    dilated=True,
+                    input_shape=tuple(self.X_train.shape[1:]))
+
+    def _set_random_seeds(self):
+        os.environ['PYTHONHASHSEED']=str(self.seed_value)
+        random.seed(self.seed_value)
+        np.random.seed(self.seed_value)
+        tf.set_random_seed(self.seed_value)
```
willgraf, line 143: Docstring would be useful. What's `PYTHONHASHSEED`?

deepcell/training/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""
+The training submodule contains all the user entry points into DeepCell's
+training functionality.
+"""
+
+#from .training import *
+from .model_trainer import ModelTrainer
```
willgraf, line 32: Add __future__ imports (and del them at the end of the file) and use absolute imports instead of relative ones to match the rest of the repo.

deepcell/training/model_trainer_test.py
```diff
@@ -0,0 +1,115 @@
...
+seed_value = 13 # value for all random seeds
+
+def _load_data():
+    # take in dataset
+    filename = 'HeLa_S3.npz'
+    test_size = 0.1 # % of data saved as test
+    (X_train, y_train), (X_test, y_test) = deepcell.datasets.hela_s3.load_data(
+                                                filename,
+                                                test_size = test_size,
+                                                seed = seed_value)
```
willgraf, line 40: Consider making some random data with numpy or some image data with skimage, rather than download a file from s3.

deepcell/training/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""
+The training submodule contains all the user entry points into DeepCell's
+training functionality.
+"""
+
+#from .training import *
```
willgraf, line 31: Import the training functions by name so that the API stays the same. (`from deepcell.training import train_model_whatever`)

deepcell/training/model_trainer_test.py
```diff
@@ -0,0 +1,115 @@
...
+# - A model [ assumed for now to be bn_feature_net_skip_2D]
+# (specified by model)
+#
+# For now, the creation of these three elements is handled in this script. They
+# are each created in separate functions, though, and those functions could be
+# easily moved over to the ModelTrainer class. As part of this, ModelTrainer
+# would need to have its  __init__ arguments updated to include all the input
+# arguments needed for creation of the element. Those decisions are all really
+# about how big we want the scope of ModelTrainer to be.
+
```
willgraf, line 22: __future__ imports to match the rest of the repo.

deepcell/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.4.0"
```
willgraf, line 1: Can we just put this in the base __init__ file? I'm not sure what is best for declaring package versions.

model_reconstitution_test.py
```diff
@@ -0,0 +1,178 @@
+# test.py
+# Test driver for the ModelTrainer class.
+#
```
willgraf, line 3: Is this file different than the one in `deepcell/training`?

setup.py
```diff
@@ -66,9 +66,10 @@ def _parse_requirements(file_path):
     logging.warning('Failed to load requirements file, using default ones.')
     install_reqs = []
 
+exec(open('deepcell/version.py').read())
 setup(
     name='DeepCell',
-    version='0.1',
+    version=__version__,
```
willgraf, line 72: Not crazy about this, are there other methods?

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+        self.trained = False
+
+        self._init_output_metadata()
+
+    def _set_random_seeds(self):
+        pass
+        #os.environ['PYTHONHASHSEED']=str(seed_value)
+        #random.seed(seed_value)
+        #np.random.seed(self.random_seed)
+        #tf.set_random_seed(self.random_seed)
```
willgraf, line 193: This should probably take a `seed` as an input.

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+        def prep_generator(
+                generator_type,
+                output_metadata,
+                data_generator,
+                input_data_dict,
+                skip = None,
+                seed = None,
+                batch_size = None,
+                transform = None,
+                transform_kwargs = {}):
```
willgraf, line 238: Honestly, I think this would be a bit simpler if `_prep_generator` was just a method instead of an internal function like this.

That way you may not need to pass so many arguments and it can directly rely on the attributes. I think it would be a bit easier to understand.

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+
+            output_metadata["generators"][generator_type_string] = {}
+            data_generator_name = re.match("<([\w.]+) object at",str(data_generator)).group(1)
+            output_metadata["generators"][generator_type_string]["class"] = data_generator_name
+            
+            parameters = {}
+            parameters["input_data_dict"] = {}
+            for entry in input_data_dict:
+                parameters["input_data_dict"][entry] = {}
+                parameters["input_data_dict"][entry]["md5_digest"] = hashlib.md5(input_data_dict[entry]).hexdigest()
```
willgraf, line 255: I think a helper function `_get_hash` or something would be nice to guarantee that it's always being hashed with the same algo. Would be eaiser to change the algo that way too if we ever wanted to do so.

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+                self.output_metadata,
+                self.model,
+                self.loss_function,
+                self.optimizer,
+                self.optimizer_passed,
+                self.optimizer_learning_rate,
+                self.optimizer_decay,
+                self.optimizer_momentum,
+                self.optimizer_nesterov,
+                metrics)
```
willgraf, line 345: I'm not sure I understand the value of the `model_compilation` function. why add this layer?

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+            # Save benchmarks in dict
+            self.benchmarks = {}
+            if self.postprocessing_fn:
+                outputs = self.model.predict(
+                    self.validation_data.x,
+                    batch_size=self.predict_batch_size)
+                y_pred = self.postprocessing_fn(outputs, **self.postprocessing_kwargs)
+                # TODO: This is a hack because the postprocessing fn returns
+                # masks with no channel dimensions. This should be fixed.
+                y_pred = np.expand_dims(y_pred, axis=-1)
```
willgraf, line 440: Maybe an if block to check if the rank is one less then expand_dims?

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+
+        # TODO: Saving the benchmarking object in this way saves each individual benchmark.
+        # This should be refactored to save the sums.
+
+        metadata_name = os.path.join(self.model_path, '{}_{}.json'.format(
+            self.model_name, self.model_hash))
+
+        with open(metadata_name, 'w') as json_file:
+            import pdb; pdb.set_trace()
+            json.dump(self.output_metadata, json_file)
```
willgraf, line 479: this could be an optional input too.

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+        self._compile_model()
+
+        # Train model with prepped data generators
+        self._train_model()
+
+        # Create benchmarks
+        self._benchmark()
+
+        # Save model
+        model_name = os.path.join(self.model_path, self.model_name + '_' + self.model_hash + '.h5')
```
willgraf, line 468: `format()` over string concatenation.

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+
+            return loss_history
+
+        def create_hash(trained, model, output_metadata):
+            if not trained:
+                raise ValueError('Can only create a hash for a trained model')
+            else:
+                weights = []
+                for layer in model.layers:
+                    weights += layer.get_weights()
```
willgraf, line 398: `.append` is much faster than `+=` for lists.

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+            self.optimizer_passed = False
+            self.optimizer_learning_rate = 0.01
+            self.optimizer_decay = 1e-6
+            self.optimizer_momentum = 0.9
+            self.optimizer_nesterov = True
+            self.optimizer = SGD(
+                    lr= self.optimizer_learning_rate,
+                    decay= self.optimizer_decay,
+                    momentum= self.optimizer_momentum,
+                    nesterov= self.optimizer_nesterov)
```
willgraf, line 171: Maybe you can save a general `kwargs` for the optimizer, instead of specifically using a `_decay` and `_momentum` attribute.

deepcell/training/model_trainer.py
```diff
@@ -0,0 +1,486 @@
...
+            """
+            Get training and validation data back from data_generator.flow() and document every step of the process
+            using output_metadata.
+            """
+
+
+            generator_type_string = generator_type + "_generator"
+
+            output_metadata["generators"][generator_type_string] = {}
+            data_generator_name = re.match("<([\w.]+) object at",str(data_generator)).group(1)
```
willgraf, line 248: this deserves a comment.

willgraf:
I'm going to close this due to inactivity. Feel free to re-open whenever it is picked up again.

PR #347, opened 2020-04-08
removed windows CRLF formatting

ngreenwald:
## What
Found file that had CRLF windows line ending character instead of unix  LF like rest of repo. I switched from the windows formatting to unix for all lines in this file. 

## Why
Can lead to diffs where everything is changed.


willgraf:
Great catch!

PR #349, opened 2020-04-14
Add preprocessing function to cytoplasm app

msschwartz21:
## What
* Add `deepcell_toolbox.processing.phase_process` as the preprocessing function for the cytoplasm application

## Why
* The model was trained using data with this preprocessing function


willgraf:
Is `phase_preprocessing` used for both phase and fluorescent cytoplasm?

msschwartz21:
> Is phase_preprocessing used for both phase and fluorescent cytoplasm?

Yes, confirmed with Dave. This PR is ready to merge.

PR #354, opened 2020-04-16
Readme update

### LINKED ISSUE #334: The README needs to be update
**Describe the bug**
The README is not clear enough for new users and some of the commands do not apply to the latest versions of Nvidia Docker. 

**The lines that are particularly concerning are:**

The fastest way to get started with DeepCell is to run the docker image:

nvidia-docker run -it --rm -p 8888:8888 vanvalenlab/deepcell-tf:0.4.0-gpu

MekWarrior:
## What
* Update several areas of the README that were incorrect or out-of-date. Add our new logo as a banner. Update the example with our latest work. Add additional funding sources.

Closes #334


deepcell/model_zoo/README.md
```diff
@@ -1,5 +1,17 @@
...
+'deepcell-tf' features several model architectures for cell segmentation. All of the models in this library are `keras.models` and can easily fit into any `keras` deep learning workflow. These include:
+
+* [FeatureNets](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005177) - This architecture was introduced in the first DeepCell paper and are quite effective for analyzing images with a single relevant length scale. This architecture is governed by a single control parameter, the receptive field, which controls the size of the spatial region around each pixel used to make predictions for that pixel. These models can perform instance segmentation by predicting a transform of the instance masks - e.g. whether pixels are edge/interior/background or a distance transform that can be used with watershed. While these are no longer the state-of-the-art, they're easy to train and are still effective - particularly for bacterial images.
```
ngreenwald, line 5:
```suggestion
* [FeatureNets](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1005177) - This architecture was introduced in the first DeepCell paper and is quite effective for analyzing images with a single relevant length scale. This architecture is governed by a single control parameter, the receptive field, which controls the size of the spatial region around each pixel used to make predictions for that pixel. These models can perform instance segmentation by predicting a transform of the instance masks - e.g. whether pixels are edge/interior/background or a distance transform that can be used with watershed. While these are no longer the state-of-the-art, they're easy to train and are still effective - particularly for bacterial images.
```

deepcell/model_zoo/README.md
```diff
@@ -1,5 +1,17 @@
...
+
+* [RetinaNet](https://arxiv.org/abs/1708.02002) - These models find bounding boxes that encapsulate cells within an image. These models only return a bounding box and do not provide an instance mask. This is effectively a port of the [fizyr retinanet library](https://github.com/fizyr/keras-retinanet) that was adapted to work with tf.keras and offer greater control of image augmentation operations. These models include the option to attach a [semantic segmentation head](http://openaccess.thecvf.com/content_CVPR_2019/html/Kirillov_Panoptic_Feature_Pyramid_Networks_CVPR_2019_paper.html) for panoptic segmentation tasks.
+
+* [RetinaMask](https://arxiv.org/abs/1901.03353) - These models use retinanet to find bounding boxes and then predict instance masks within each bounding boxes. These models are essentially the same as the popular [MaskRCNN](https://arxiv.org/abs/1703.06870) method, except it uses RetinaNet to generatte the bounding boxes as opposed to FasterRCNN. This is effectively a port of the [fizyr maskrcnn library](https://github.com/fizyr/keras-maskrcnn). These models include the option to attach a semantic segmentation head.
```
ngreenwald, line 11:
```suggestion
* [RetinaMask](https://arxiv.org/abs/1901.03353) - These models use retinanet to find bounding boxes and then predict instance masks within each bounding boxes. These models are essentially the same as the popular [MaskRCNN](https://arxiv.org/abs/1703.06870) method, except it uses RetinaNet to generate the bounding boxes as opposed to FasterRCNN. This is effectively a port of the [fizyr maskrcnn library](https://github.com/fizyr/keras-maskrcnn). These models include the option to attach a semantic segmentation head.
```

README.md
```diff
@@ -41,26 +43,31 @@ Tracked Image
 
 ## Getting Started
```
ngreenwald, line 44: I think we should mention that all of this currently requires a GPU. So far reading this I wouldn't know that this

README.md
```diff
@@ -41,26 +43,31 @@ Tracked Image
...
 
-DeepCell uses `nvidia-docker` and `tensorflow` to enable GPU processing. If using GCP, there are [pre-built images](https://console.cloud.google.com/marketplace/details/nvidia-ngc-public/nvidia_gpu_cloud_image) which come with CUDA, docker, and nvidia-docker pre-installed. Otherwise, you will need to install [docker](https://docs.docker.com/install/linux/docker-ce/debian/), [nvidia-docker](https://github.com/NVIDIA/nvidia-docker), and [CUDA](https://developer.nvidia.com/cuda-downloads) separately.
+`deepcell-tf` uses `nvidia-docker` and `tensorflow` to enable GPU processing. If using GCP, there are [pre-built images](https://console.cloud.google.com/marketplace/details/nvidia-ngc-public/nvidia_gpu_cloud_image) which come with CUDA, docker, and nvidia-docker pre-installed. Otherwise, you will need to install [docker](https://docs.docker.com/install/linux/docker-ce/debian/), [nvidia-docker](https://github.com/NVIDIA/nvidia-docker), and [CUDA](https://developer.nvidia.com/cuda-downloads) separately. These instructions set up a docker container so that you can directly change the `deepcell-tf` library itself and have those changes reflected within the container
```
ngreenwald, line 70: Would it make more sense to include this information about pre-built instances in the getting started section?

If a relatively novice person tries to get this up and running, they're going to get sent to the nvidia-docker link in the getting started section and instantly get stuck in CUDA library dependency mess.

deepcell/model_zoo/README.md
ngreenwald: Hmm, this didn't show up very well. Deepcell paper are -> Deepcell paper is

deepcell/model_zoo/README.md
ngreenwald: generatte -> generate

README.md
MekWarrior: it doesn't require a GPU, there is a CPU version

ngreenwald: Yeah definitely. If you think it's not worth mentioning the performance difference between GPU and CPU then no need to include

README.md
MekWarrior: I see @ngreenwald's point here. The historical context is that we have always introduced new users to deepcell-tf under the assumption that they had access to a server where all the standard dependencies are already in place (Docker, Jupyter, and all the correct drivers). While we've had users who prefer to get started with VMs, that has not been our preferred introduction. Additionally, our docs are structured in such a way as to provide minimal starting information on the READMEs with a pointer to RTD. This was a design decision to keep the READMEs as simple as possible. That being said, both are still under construction and in need of improvements.

I'd welcome input from @willgraf and @msschwartz21 before proceeding

msschwartz21: I agree with @ngreenwald that our existing Getting Started section makes several non-trivial assumptions about the user's local environment and their access to GPU. I also think that we should still continue to prioritize keeping READMEs as simple as possible.

Perhaps as a compromise, we can make the following changes starting at line 46:
> The fastest way to get started with `deepcell-tf` is to run one of our existing images from Docker Hub. To do so locally, you need access to a GPU with [docker](https://docs.docker.com/install/linux/docker-ce/debian/), [nvidia-docker](https://github.com/NVIDIA/nvidia-docker), and [CUDA](https://developer.nvidia.com/cuda-downloads). Alternatively, Google Cloud Platform (GCP) offers [prebuilt virtual machines](https://console.cloud.google.com/marketplace/details/nvidia-ngc-public/nvidia_gpu_cloud_image) preinstalled with Cuda, docker and nvidia-docker.

Then in the developer docs we can just focus on building a docker image for development.

willgraf: We currently maintain 2 images for each release, the CPU and GPU builds.

Maybe we can introduce the "easiest way" as using the CPU build, which should run anywhere, and we can have another section about GPU usage where we go into more detail about `nvidia-docker` and on-prem vs cloud GPU servers.

README.md
```diff
@@ -41,26 +43,31 @@ Tracked Image
...
+# Start a GPU enabled container on one GPUs
+docker run --gpus '"device=0"' -it --rm \
+	-p 8888:8888 \
+    -v $PWD/scripts:/notebooks \
+    -v $PWD/data:/data \
+    vanvalenlab/deepcell-tf:0.4.0-gpu
 ```
 
-This will start a Jupyter session, with several example notebooks detailing various training methods:
+This will spin up a docker container with `deepcell-tf` installed and start a jupyter session using the default port 8888. This command also connects a data folder ($PWD/data) and a scripts folder ($PWD/scripts) to the docker container so it can access your data and juyter notebooks stored on the host workstation. The default port can be changed to any non-reserved port by updating `-p 8888:8888` to, e.g., `-p 81:8888`. If you run across any errors getting started, you should either refer to the `deepcell-tf` for developers section or raise an issue on GitHub.
```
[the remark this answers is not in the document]
willgraf: I think the blurb "so it can access your data and notebooks" does a lot of that explanation already. Maybe just a sentence like "any saved data or models must be saved in the mounted directory to be accessible outside of the container" would suffice?

 not explain clearly enough that the container will not be able to access without it?

Additionally, we should change the word "connects" to "mounts" to use more google-able words.

README.md
MekWarrior: Added mention of CPU and VM alternatives in "Getting Started"

README.md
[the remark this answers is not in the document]
MekWarrior: Done

PR #352, opened 2020-04-15
Update panopticnet and fpn for future model export support

msschwartz21:
## What
- Removed TensorProduct for Conv's in PanopticNet
- Added names to all the layers of PanopticNet (Closes #338)
- Removed semantic_prediction which wasn't being used in the file
- Added an option to use depthwise conv's in feature pyramid construction as opposed to regular convs

## Why
* TensorProduct not supported for TfLite
* Names required for saved model export

PR #355, opened 2020-04-17
Update weights url and hashes for nuclear and cytoplasm models

msschwartz21:
## What
* Update models with new weights url and md5 hash


msschwartz21:
These weights files are dependent on #352. We should hold off on merging this PR until #352 has been merged into master.

willgraf:
Are these weights files trained after the modifications to `panopticnet.py`? Specifically the `TensorProduct` removal and the named layers?

msschwartz21:
> Are these weights files trained after the modifications to panopticnet.py? Specifically the TensorProduct removal and the named layers?

Yes I think so. This PR was dependent on the changes from #352, which I think includes the changes you referenced above.

PR #356, opened 2020-04-18
Update centroid transform so it scales with cell size

vanvalen:
Added auto_alpha to centroid_transform_continuous_2d so that it scales with cell size

## What
* The centroid transform transforms each pixel inside the cell to 1/(1+alpha*x) where x is the distance to the center of mass in pixels and alpha is a parameter that sets a length scale. This alpha is fixed, which means that it is the same for every cell. This is problematic for big cells. In this pull request, we change alpha to 1/sqrt(cell area) so that it scales with cell size.

## Why
* Enables more accurate segmentation of cytoplasmic images


deepcell/image_generators/__init__.py
```diff
@@ -174,6 +174,9 @@ def _transform_masks(y, transform, data_format=None, **kwargs):
     elif transform == 'centroid':
         erosion = kwargs.pop('erosion_width', 0)
         disk_size = kwargs.pop('disk_size', 4)
+        alpha = kwargs.pop('alpha', 0.1)
+        auto_alpha = kwargs.pop('auto_alpha', False)
```
willgraf, line 178: `auto_alpha` Got removed from the function, so it can be removed here.

PR #370, opened 2020-04-24
PanopticNet Updates

### LINKED ISSUE #367: New layer names breaks deep watershed losses construction
**Describe the bug**
Switching the names of the layers in the deep watershed model to include 'semantic' means that they get picked up by the loss construction logic, resulting in too many losses. 

**To Reproduce**
Run the Deep Watershed example notebook. 

**Screenshots**
image

**Suggested fix**
We can either change the names back to not include 'semantic', or modify the following code to be more specific and not grab everything with semantic in the name
```
# Give losses for all of the semantic heads
for layer in model.layers:
    print(layer.name)
    if 'semantic_' in layer.name:
        n_classes = layer.output_shape[-1]
        loss[layer.name] = semantic_loss(n_classes)
```

### LINKED ISSUE #368: Panopticnet models don't build with non-square inputs
**Describe the bug**
When trying to create a panopticnet an operand broadcast error gets thrown

**To Reproduce**
Try and run the deep watershed example notebook. 

**Screenshots**
image

**Additional context**
I didn't notice this earlier because it only gets thrown in particular circumstances. 

- if the input image crop is square it doesn't throw an error (this is how it slipped past the [test suite](https://github.com/vanvalenlab/deepcell-tf/blob/master/deepcell/model_zoo/panopticnet_test.py#L83): 
- If the input dimensions are sufficiently small (errors on 216x256, not 28x30)

Looks like it was introduced in #352

ngreenwald:
## What
Updates example notebook with new loss generator code, adds checks for invalid input sizes. 

Closes #367, closes #368


willgraf:
Is this another line-ending character issue?

ngreenwald:
Yeah, looks like same thing. Let me see if I can just pass it back with the windows character. 

Do you want me to add a test with a bad input size to address the coverage issue?

willgraf:
> Do you want me to add a test with a bad input size to address the coverage issue?

Yes please!

deepcell/model_zoo/panopticnet.py
```diff
@@ -326,6 +326,15 @@ def PanopticNet(backbone,
...
                                                str(acceptable_modes)))
 
+    
+    # TODO only works for 2D: do we check for 3D as well? What are the requirements for 3D data?
+    img_shape = input_shape[1:] if channel_axis == 1 else input_shape[:-1]
+    if img_shape[0] != img_shape[1]:
+        raise ValueError('Input data must be square, got dimensions {}'.format(img_shape))
+
+    if img_shape[0] not in [2 ** x for x in range(5, 12)]:
+        raise ValueError('Input data dimensions must be a power of 2, got {}'.format(img_shape[0]))
```
willgraf, line 336: Why not just use `log` to check? That would remove the need for the size limits and would be O(1) instead of O(N).

```python
if math.log(img_shape[0], 2).is_integer():
  ...
```

ngreenwald:
Alright, let me know what you think. I notice some of the machinery is in place for 3D predictions, but the test functions don't test 3D, so I wasn't sure what to do for the input_size validation

deepcell/model_zoo/panopticnet_test.py
```diff
@@ -112,3 +112,45 @@ def test_panopticnet(self, pooling, location):
...
+                    backbone=backbone,
+                    input_shape=input_shape,
+                    backbone_levels=['C3', 'C4', 'C5'],
+                    norm_method=norm_method,
+                    location=True,
+                    pooling='avg',
+                    num_semantic_heads=len(num_semantic_classes),
+                    num_semantic_classes=num_semantic_classes,
+                    use_imagenet=False,
+                )
```
willgraf, line 156: This should be the same scope as the first `with self.assertRaises` call (it's indented once too many).

ngreenwald:
Nice, I like the log check, that's clever. Pushed those two

PR #371, opened 2020-04-27
Multiplex application

ngreenwald:
## What
* Add in an application for multiplex segmentation
* Modify existing application setup to use new version of resize which preserves labels

## Why
* Adds ability to use current version of multiplex model


requirements.txt
```diff
@@ -12,4 +12,4 @@ opencv-python>=3.4.2.17,<4
 cython>=0.28
 pathlib==1.0.1
 deepcell-tracking>=0.2.4
-deepcell-toolbox>=0.3.0
+git+https://github.com/vanvalenlab/deepcell-toolbox.git@resize_labels
```
willgraf, line 15: This has to be in a real release before we can merge this PR.

ngreenwald:
Good catch, update all the docstrings to include new ValueError

deepcell/applications/application.py
```diff
@@ -240,16 +242,13 @@ def _resize_output(self, image, original_shape):
...
             new_shape = original_shape[1:-1]
             # Flip order of shape axes to prevent transpose of data
             new_shape = new_shape[::-1]
-            image = resize(image.astype('float32'),
-                           new_shape, data_format='channels_last')
-            image = image.astype(intype)
-
+            image = resize(image,
+                           new_shape, data_format='channels_last',
+                           labeled_image=True)
```
willgraf, line 251:
```suggestion
            image = resize(image, new_shape,
                           data_format='channels_last',
                           labeled_image=True)
```

deepcell/applications/application.py
```diff
@@ -300,10 +306,10 @@ def _predict_segmentation(self,
         output_tiles = self.model.predict(tiles, batch_size=batch_size)
 
         # Untile images
-        output_images = self._untile_output(output_tiles, tiles_info)
+        self.output_images = self._untile_output(output_tiles, tiles_info)
 
         # Postprocess predictions to create label image
-        label_image = self._postprocess(output_images, **postprocess_kwargs)
+        label_image = self._postprocess(self.output_images, **postprocess_kwargs)
```
willgraf, line 312: I am not a big fan of creating attributes outside of `__init__`. Is this required?

deepcell/applications/multiplex_segmentation.py
```diff
@@ -0,0 +1,175 @@
...
+        Returns:
+            np.array: Labeled image
+            np.array: Model output
+        """
+        return self._predict_segmentation(image,
+                                          batch_size=batch_size,
+                                          image_mpp=image_mpp,
+                                          preprocess_kwargs=preprocess_kwargs,
+                                          postprocess_kwargs=postprocess_kwargs), \
+            self.output_images
```
willgraf, line 175: I'm not sure about returning output_images at all, but this is hard to read regardless.

```suggestion
        prediction = self._predict_segmentation(
            image,
            batch_size=batch_size,
            image_mpp=image_mpp,
            preprocess_kwargs=preprocess_kwargs,
            postprocess_kwargs=postprocess_kwargs)
        return  prediction, self.output_images
```

ngreenwald: The other option would be not to return the output in predict, and just to store it as self.variable, which can then be accessed directly?

Basically, I want the option to be able to re-run the post-processing function with different parameters on the raw images, since currently this model often requires some manual tweaking. I'm not sure what the best way to get at the underlying model output is, totally open to suggestions

willgraf: I see. Maybe we could add another method to the base `Application` class (something like `_predict`) that does all the tiling/untiling and returns the model output. Then change `_predict_segmentation` to use the new method.

ngreenwald: For sure. @msschwartz21 and I had discussed maybe adding a multiplex-specific segmentation function. Any thoughts on this? Happy to go with whatever you both think makes the most sense.

willgraf: How would it differ?

ngreenwald: Ah got it, okay that makes sense. So one function would return the output. A second function would produce the post-processed model output, and would call the first function/check if the model output has already been saved as internal variable?

willgraf: I think the second function can just take the output as an argument and return the results. (it may exist already as `_postprocess` or something). The outer wrapper for `predict` can just call both.  I'm honestly not sure that `_predict_segmentation` should not just be the body of `predict`, but that's kind of another topic.

I don't think we need an attribute for the output data. (but I could be convinced if there's a good argument for it).

ngreenwald: Yeah, I also wasn't super clear on the benefit of the _predict_segmentation/predict dichotomy. 

Okay, that sounds good, I'll make those changes. As long as there's some way to directly return model output I don't think it needs to be saved as an attribute.

msschwartz21: I originally pulled `_predict_segmentation` out as a separate function from `predict` to allow future flexibility for models with a different workflow, e.g. tracking.

ngreenwald: Ah got it, so the tracking app would still call predict, but it wouldn't simply be a case of a different post-processing function, it would be a fundamentally different workflow?

willgraf: I think it's desired that the main interface for the application is `app.predict()`.

Each one can have a different `predict` implementation, but they will likely re-use a bunch of pieces.

ngreenwald:
Alright, let me know what you guys think of the current reorg

deepcell/applications/application.py
```diff
@@ -277,35 +308,31 @@ def _predict_segmentation(self,
...
 
         # Check input size of image
         if len(image.shape) != self.required_rank:
-            raise ValueError('Input data must have {} dimensions'
+            raise ValueError('Input data must have {} dimensions. '
                              'Input data only has {} dimensions'.format(
                                  self.required_rank, len(image.shape)))
 
-        # Resize image, returns unmodified if appropriate
-        image, original_shape = self._resize_input(image, image_mpp)
```
msschwartz21, line 291: I don't like having the first resize hidden inside `_run_model` while the second is exposed in `_predict_segmentation`. I would vote for keeping both resize steps in `_predict_segmentation`.

ngreenwald:
Okay, I pulled out the resize function. 

Perhaps this is a conversation for another time, but I wonder if it would be worth separating what tests are run against all builds and what are only run in PRs against master? I think we could do something like that with [custom pytest markers](https://docs.pytest.org/en/latest/example/markers.html) and [conditional travis builds](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/). 

These application tests take a longggg time to run, since we're doing inference on a CPU with these massive models.

willgraf:
> These application tests take a longggg time to run, since we're doing inference on a CPU with these massive models.

I do not think the applications need to be doing inference at all. Testing the methods does not need to test the `PanopticModel` too.  My suggestion to fix the long-running application tests is to create an `Application._get_model` helper method that we can monkey patch during testing so that we do not need to do any actual inference.

ngreenwald:
Sure thing. Do you want me to do that in this PR?

willgraf:
No need, I'll make another PR soon. The applications need to be updated to drop the fgbg head as well.

requirements.txt
```diff
@@ -12,4 +12,4 @@ opencv-python>=3.4.2.17,<4
 cython>=0.28
 pathlib==1.0.1
 deepcell-tracking>=0.2.4
-deepcell-toolbox>=0.4.0
+deepcell-toolbox>=0.4.0
```
willgraf, line 15: What changed here?

PR #373, opened 2020-04-30
Correct calculation of image size for rescale

msschwartz21:
## What
* Correct calculation of image size for rescale in application objects
## Why
* The original calculation was the inverse of what it should have been


msschwartz21:
We were testing the calculation in a way, but I had it set up wrong so the test was passing with the original conditions. Should be correct now.

PR #379, opened 2020-05-05
fixed resize bug

### LINKED ISSUE #378: Applications dont rescale outputs
**Describe the bug**
If the mpp is different than the training data, the applications objects don't rescale the outputs.

**To Reproduce**
Run application object with a custom mpp

ngreenwald:
Fixes bug that resulted in wrong shape being passed to resize. Also adds new test to check that resize works appropriately, and one to check that post-processing function runs as expected. 

Closes #378

### DIFF
--- a/deepcell/applications/application_test.py
+++ b/deepcell/applications/application_test.py
@@ -74,24 +74,20 @@ class TestApplication(test.TestCase):
         x = np.random.rand(1, 500, 500, 1)
 
         # image_mpp = None --> No resize
-        y, original_shape = app._resize_input(x, image_mpp=None)
+        y = app._resize_input(x, image_mpp=None)
         self.assertEqual(x.shape, y.shape)
-        self.assertEqual(x.shape, original_shape)
 
         # image_mpp = model_mpp --> No resize
-        y, original_shape = app._resize_input(x, image_mpp=kwargs['model_mpp'])
+        y = app._resize_input(x, image_mpp=kwargs['model_mpp'])
         self.assertEqual(x.shape, y.shape)
-        self.assertEqual(x.shape, original_shape)
 
         # image_mpp > model_mpp --> resize
-        y, original_shape = app._resize_input(x, image_mpp=2.1 * kwargs['model_mpp'])
+        y = app._resize_input(x, image_mpp=2.1 * kwargs['model_mpp'])
         self.assertEqual(2.1, np.round(y.shape[1] / x.shape[1], decimals=1))
-        self.assertEqual(x.shape, original_shape)
 
         # image_mpp < model_mpp --> resize
-        y, original_shape = app._resize_input(x, image_mpp=0.7 * kwargs['model_mpp'])
+        y = app._resize_input(x, image_mpp=0.7 * kwargs['model_mpp'])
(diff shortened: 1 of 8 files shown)


deepcell/applications/application_test.py
```diff
@@ -165,6 +167,14 @@ def _postprocess(Lx):
...
         self.assertAllEqual(np.ones(x.shape), y)
 
+        # deep watershed
+        model_outputs = [np.zeros((3, 30, 30, 1)), np.zeros((3, 30, 30, 1)),
+                         np.zeros((3, 30, 30, 1))]
+        kwargs = {'postprocessing_fn': deep_watershed}
+        app = Application(model, **kwargs)
+        y = app._postprocess(model_outputs)
+        self.assertEqual(y.shape, model_outputs[0].shape)
+
```
willgraf, line 177: Is there a way to use a dummy post-processing step on this? I don't think we need to require functions outside of deepcell to test the code.

ngreenwald: For sure. If you look right above the new changes, you can see that Morgan already has such a dummy function. I thought adding this in might be good to catch any inadvertent changes that would break the post-processing, but if you think it's not necessary I can remove it

willgraf: I think the dummy meets the testing requirement, but you're welcome to include a test if it addresses another facet of post-processsing. My main concern is that it's testing `deep_watershed` more than it's testing the application itself.

deepcell/applications/application_test.py
```diff
@@ -232,3 +242,11 @@ def test_predict_segmentation(self):
...
         y = app._predict_segmentation(x)
         self.assertEqual(x.shape, y.shape)
+
+        # test with different MPP
+        model = DummyModel()
+        app = Application(model)
+
+        x = np.random.rand(1, 128, 128, 1)
+        y = app._predict_segmentation(x, image_mpp=1.3)
+        self.assertEqual(x.shape, y.shape)
```
willgraf, line 252: This is the test that addresses the original issue, right?

ngreenwald: Yes, exactly

deepcell/applications/application_test.py
ngreenwald: I guess what I was trying to address with this test is that our post-processing functions currently expect the model output to be a list of arrays, whereas the tests in the app for post-processing instead just pass a single array. But given that different post-processing functions could expect different outputs, it's probably not worth encoding that in the app itself?

willgraf: Maybe we can write a test for another dummy function that uses an array of length 2.

That way we can kinda test both lists and arrays generally for `Application`. That way subclasses can test their own post processing.

deepcell/applications/application_test.py
```diff
@@ -36,6 +36,8 @@
 
 from deepcell.applications import Application
 
+from deepcell_toolbox.deep_watershed import deep_watershed
+
```
willgraf, line 40:
```suggestion
```

ngreenwald: For sure, I think for testing that the pipeline itself hands off correctly the one that's there covers the bases.

PR #384, opened 2020-05-06
Harmonize semantic heads

### LINKED ISSUE #363: Harmonize semantic heads between fpn.py and panoptic.py
We currently have two different versions of semantic heads - one in fpn and panoptic nets. It would be nice to harmonize them so we only have one set.

One set of semantic heads used across the code base.

willgraf:
## What
* Merge the FPN functions from `model_zoo/panopticnet.py` into `model_zoo/fpn.py`.

## Why
* Having duplicated / slightly different code is hard to manage. This PR unifies the functionality of the feature pyramid networks. 
* Fixes #363


deepcell/model_zoo/fpn.py
```diff
@@ -461,104 +488,52 @@ def __create_semantic_head(pyramid_dict,
...
+
+    semantic_sum = pyramid_features[-1]
 
-    for N, P in zip(pyramid_names, pyramid_features):
-        # Get level and determine how much to upsample
-        level = int(re.findall(r'\d+', N)[0])
+    # Final upsampling
+    min_level = int(re.findall(r'\d+', pyramid_names[-1])[0])
+    # n_upsample = min_level - target_level
+    n_upsample = min_level
```
willgraf, line 517: I was unsure about this.

willgraf: Both `RetinaNet(panoptic=True)` and `PanopticNet` create semantic heads with `target_level` calculated as:

```python
pyramid_dict = create_pyramid_features(...)
semantic_levels = [int(re.findall(r'\d+', k)[0]) for k in pyramid_dict]
target_level = min(semantic_levels)
```

With the current code above, we ignore `target_level` and calculates `n_upsample` for the final semantic prediction without `target_sample`.

```python
min_level = int(re.findall(r'\d+', pyramid_names[-1])[0])
# n_upsample = min_level - target_level
n_upsample = min_level
```

I think that `target_level` will always be equal to `min_level`. Should the final n_upsample usually be `0`? If not, can we just shortcut to `n_upsample = target_level`?

willgraf:
Approved via Slack.

PR #389, opened 2020-05-08
metrics module is moved to deepcell_toolbox.metrics.

willgraf:
Leaving imports in metrics.py to keep the API the same.

## What
* Remove code from `metrics.py` and replace with imports from `deepcell_toolbox.metrics`.
* Remove `metris_test.py` entirely.

## Why
* The metrics package has been migrated to `deepcell_toolbox` and there is no reason to have the duplicate code.


willgraf:
Coverage dropping 1% isn't great but we removed a lot of lines of code. 

> "Progress takes sacrifice"

PR #397, opened 2020-05-15
Added missing name in PanopticNet BatchNormalization layer

willgraf:
## What
* Added a name to the batch norm layer in PanopticNet.
* Updated weights files for Applications.

## Why
* Enables models to be loaded by name to easily remove unused semantic heads.


willgraf:
Confirmed all application weights are loaded and run with data.

PR #406, opened 2020-06-09
random cropping for image generators

ngreenwald:
## What
Adds in the option to specify a crop_size, which will be subsampled randomly from the provided image

## Why
Useful data augmentation strategy that alleviates the need for us to generate overlapping crops as a preprocessing step, reducing the size of the trained NPZ as well as redundant precomputing of transforms

## Current issues
For some reason I kept getting an "unexpected keyword arugment" error when overriding interpolation order in the ImageDataGenerator __init__ function, so I commented it out for now. 

The fit() method also currently fails with cropping, since I didn't override that method from Keras. Happy to add it in, but I don't currently use any of the transform that require fit() to runs, so I left it out for now to get initial feedback


ngreenwald:
Just tried this out, it's awesome. @willgraf , let me know how you want me to go about addressing the Keras fit() issue, and if you have any ideas about interpolation order

ngreenwald:
Hey @willgraf, I modified the last commit to reproduce the error. It looks like the ImageDataGenerator we're inheriting from isn't expecting the interpolation_order as an argument to __init__. I'm not sure why that is, since based on looking at the code on their github it seems to be defined. But the fairly easy workaround is what I did in the previous commit, which is just defining it in our version of __init__ and not passing it through to the super __init__

willgraf:
Oh interesting, it looks like `interpolation_order` is supported in [keras-preprocessing](https://github.com/keras-team/keras-preprocessing/blob/422f89057f930f61086daf028ccea0ab8eab71ae/keras_preprocessing/image/image_data_generator.py#L276) but when that class is imported over overridden in tensorflow, [they do not include interpolation_order](https://github.com/tensorflow/tensorflow/blob/fbe175972d6c9fe354068f2d51f6a357e414c77c/tensorflow/python/keras/preprocessing/image.py#L753).

I think creating our own init class is a good solution. We may want to add a comment explaining the situation and why we're overriding the init.

ngreenwald:
Okay great, just made that change. Let me know what you think about the rest

deepcell/image_generators/__init__.py
```diff
@@ -254,6 +254,9 @@ def _transform_masks(y, transform, data_format=None, **kwargs):
 
 from deepcell.image_generators.tracking import SiameseDataGenerator
 from deepcell.image_generators.tracking import SiameseIterator
+
+from deepcell.image_generators.cropping import CroppingDataGenerator
+from deepcell.image_generators.cropping import CropperIterator
```
willgraf, line 259: The prefixes for these two are different (`Cropping` and `Cropper`). Let's make them the same name, for easier importing etc.

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
+from __future__ import absolute_import
```
willgraf, line 1: Copy/paste the copyright from another file.

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
...
+        self.y_semantic_list = []  # optional semantic segmentation targets
+
+        # set output size based on cropping or not
+        if crop_size is not None:
+            output_size = crop_size
+
+        else:
+            output_size = self.x.shape[1:3] if self.channel_axis == 3 else self.x.shape[2:4]
+
+        self.output_size = output_size
```
willgraf, line 106: Is this all that is different from the parent class `init`?

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
...
+                            'objects.', invalid_batches.size, self.x.shape[0],
+                            self.min_objects)
+
+        self.x = np.delete(self.x, invalid_batches, axis=0)
+        self.y = np.delete(self.y, invalid_batches, axis=0)
+        self.y_semantic_list = [np.delete(y, invalid_batches, axis=0)
+                                for y in self.y_semantic_list]
+
+        super(SemanticIterator, self).__init__(
+            self.x.shape[0], batch_size, shuffle, seed)
```
willgraf, line 148:
```suggestion
        super(CropperIterator, self).__init__(
            self.x.shape[0], batch_size, shuffle, seed)
```

call `super` on the current class to run the parent class' `init` method. I think you can probably remove most of the above code and call the `SemanticIterator.__init__` right here instead.

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
...
+                                    channel_shift_range=channel_shift_range,
+                                    fill_mode=fill_mode,
+                                    cval=cval,
+                                    horizontal_flip=horizontal_flip,
+                                    vertical_flip=vertical_flip,
+                                    rescale=rescale,
+                                    preprocessing_function=preprocessing_function,
+                                    data_format=data_format,
+                                    validation_split=validation_split,
+                                    dtype=dtype)
```
willgraf, line 335: What are you trying to do here? You can call the parent class' init method with:

```
super(CroppingDataGenerator, self).__init__(...)
```

What you currently have does not change the state of the class (I don't think). The `ImageDataGenerator` is not linked to `self`.

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
...
+            seed=seed,
+            data_format=self.data_format,
+            save_to_dir=save_to_dir,
+            save_prefix=save_prefix,
+            save_format=save_format,
+            crop_size=self.crop_size)
+
+    def get_random_transform(self, img_shape, seed=None):
+        transform_parameters = ImageDataGenerator.get_random_transform(self, img_shape=img_shape,
+                                                                       seed=seed)
```
willgraf, line 397: Instead, you can use `super(CroppingDataGenerator, self).get_random_transform` (`CroppingDataGenerator` is a `SemanticGenerator` which is a `ImageDataGenerator`).

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
...
+        if transform_parameters['crop_indices'] is not None:
+            row_indices, col_indices = transform_parameters['crop_indices']
+            if self.channel_axis == 1:
+                x = x[:, row_indices[0]:row_indices[1], col_indices[0]:col_indices[1]]
+            else:
+                x = x[row_indices[0]:row_indices[1], col_indices[0]:col_indices[1], :]
+
+        x = ImageDataGenerator.apply_transform(self,
+                                               x=x,
+                                               transform_parameters=transform_parameters)
```
willgraf, line 432: here too, use the super class' method.

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
...
+        return x
+
+    def random_transform(self, x, y=None, seed=None):
```
willgraf, line 435: Is it necessary to override the parent class' version?

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,576 @@
...
+                    self.interpolation_order = 0
+                y_new.append(y_t)
+            y = y_new
+        else:
+            y = self.apply_transform(y, params)
+
+        self.interpolation_order = _interpolation_order
+        return x, y
+
+    def fit(self, x, augment=False, rounds=1, seed=None):
```
willgraf, line 485: Is it necessary to override the parent class' version?

deepcell/image_generators/cropping.py
ngreenwald: So the SemanticDataGenerator doesn't have an __init__ defined. However, I need to initialize this self.crop_size parameter for the cropping generator. 

I assumed that I had to call __init__ with the parameters passed in, which is why I did that way. But since we're not actually changing any of the defaults I'm guessing your version would work, which is much simpler

deepcell/image_generators/cropping.py
ngreenwald: Is that syntactically more clear than ImageDataGenerator.get_random_transform? 
Because then its obvious that its a superclass. The way its currently written there's no way to know its the parent class, right?

deepcell/image_generators/cropping.py
ngreenwald: Hmm, I thought so, but looks like not

deepcell/image_generators/cropping.py
ngreenwald: Yes, lines 531-535 are new

deepcell/image_generators/cropping.py
willgraf: Ah, but the `SemanticDataGenerator` inherits an `init` method from its parent class, `ImageDataGenerator`. The way you have it now is not initializing the parent class.

Since you have to override the init method at all, we can just initialize the parent class first, then add your custom attributes. We still need to pass all those default values through though.

deepcell/image_generators/cropping.py
willgraf: Yes, it groups all the logic in a more clear way. Functionally, it's better because it keeps our functionality consistent.  If we went ahead and changed the `SemanticDataGenerator` in some way that overrides the `get_random_transform`, we want to be able to maintain those changes.

Using `super` just more naturally utilizes the object ineheritance.

deepcell/image_generators/cropping.py
willgraf: I think there is a test function we can add to the test suite for the `fit` method. Try looking at the MovieDataGenerator tests.

ngreenwald: I didn't see anything beyond just calling fit, which the code currently does

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,493 @@
...
+            data_format=self.data_format,
+            save_to_dir=save_to_dir,
+            save_prefix=save_prefix,
+            save_format=save_format,
+            crop_size=self.crop_size)
+
+    def get_random_transform(self, img_shape, seed=None):
+        transform_parameters = \
+            super(CroppingDataGenerator, self).get_random_transform(img_shape=img_shape,
+                                                                    seed=seed)
```
willgraf, line 365:
```suggestion
        transform_parameters = super(CroppingDataGenerator, self).get_random_transform(
            img_shape=img_shape, seed=seed)
```

The `\` is bothering me...Does this look any better?

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,493 @@
...
+
+        if transform_parameters['crop_indices'] is not None:
+            row_indices, col_indices = transform_parameters['crop_indices']
+            if self.channel_axis == 1:
+                x = x[:, row_indices[0]:row_indices[1], col_indices[0]:col_indices[1]]
+            else:
+                x = x[row_indices[0]:row_indices[1], col_indices[0]:col_indices[1], :]
+
+        x = super(CroppingDataGenerator,
+                  self).apply_transform(x=x, transform_parameters=transform_parameters)
```
willgraf, line 399:
```suggestion
        x = super(CroppingDataGenerator, self).apply_transform(
            x=x, transform_parameters=transform_parameters)
```

Same here, personal preference but I think this is a bit easier to read.

deepcell/image_generators/cropping.py
```diff
@@ -0,0 +1,493 @@
...
+    def fit(self, x, augment=False, rounds=1, seed=None):
+        """Fits the data generator to some sample data.
+        This computes the internal data stats related to the
+        data-dependent transformations, based on an array of sample data.
+        Only required if `featurewise_center` or
+        `featurewise_std_normalization` or `zca_whitening` are set to True.
+        When `rescale` is set to a value, rescaling is applied to
+        sample data before computing the internal data stats.
+        # Arguments
+            x: Sample data. Should have rank 4.
```
willgraf, line 411:
```suggestion
        sample data before computing the internal data stats.

        Args:
            x: Sample data. Should have rank 4.
```

ngreenwald:
Sure thing, done!

ngreenwald:
Cool, also added a test for crop_size=image size to get line 372

Issue #403, opened 2020-06-09
Memory/speed issues during image_generator label preprocessing


ngreenwald:
**Is your feature request related to a problem? Please describe.**
The current image generators are fairly memory intensive, and also pretty slow. The speed issue would be nice to improve, but not as crucial as the memory issues. In particular, the most recent dataset that I tried to train on resulted in a memory error on GCP. I'm going to upgrade from n1-standard-8 to n1-highmem-8 as an interim fix, but given that our dataset sizes are going to increase dramatically moving forward, we won't be able to upgrade our way out of this.

**Describe the solution you'd like**
It looks like the transforms are currently all using [float64](https://github.com/vanvalenlab/deepcell-tf/blob/0900be77f999bc514a47f92da774bb5fd35c8df1/deepcell/image_generators/__init__.py#L107) (numpy defaults) when creating transforms. Switching to float16 for the transforms would reduce the memory burden substantially.

Currently the NPZ is significantly larger due to the semi-overlapping crops I've created. Adding functionality to the generator to sub-sample crops of a given size from the images present in the NPZ would alleviate this. 

Currently, the transforms for entire dataset are pre-computed in memory. Is there a way this can be broken up into smaller steps, perhaps in a way that would allow parallel processing?

ngreenwald:
The memory issues appear to be solved now with the merged and pending PRs. 

However, I'm wondering what the best approach is for the speed issue. Currently, with 230k cells, it takes a little over an hour to instantiate the DataGenerators. As the number of annotations continues to increase this is going to become prohibitive.

Issue #382, opened 2020-05-06
Out of memory errors - change data generators


esgomezm:
**Is your feature request related to a problem? Please describe.**
The notebooks provided in ` /deepcell-tf/scripts/ ` cannot be easily used in Google Colab due to the size of the data employed during the training. I see two main problems:
- At the beginning of the notebook, when the training and test datasets are loaded, especially with `HeLa_S3.npz`. The dataset is loaded just to fix the `input_shape` of the network. 
- During the training, the entire dataset is loaded and augmented, which cannot be supported by a common RAM memory. This also happens in our current server, so I needed to modify the way the data is loaded during the training.

**Describe the solution you'd like**
- Determine the input shape without loading 6500 images. Especially when these images will not be used for the training routine as that is done in the background, so it implies a waste of memory. 
- Change the training data generator to avoid having 5M of images in memory.

**Describe alternatives you've considered**
- The images in `HeLa_S3.npz` seem to be already patches of a larger image, cropped in a random way. Hence it seems reasonable to define a data generator that reads the images from a directory and crops a certain number of augmented patches on each epoch so it does not saturate the RAM memory and the process can be incorporated into the training workflow. Otherwise, I would reduce the number of augmented patches and increase the number of training epochs.

PR #412, opened 2020-07-03
Use smaller default dtypes

ngreenwald:
## What
* Modified the SemanticDataGenerator to use significantly less memory.
* Modified the _transform_masks function to return smallest possible dtype when computing transformations
* Added in warning message when input data is greater than precision16, as this accounts for the bulk of the memory footprint during preprocessing
* Set default representation for all data to be precision16
* Cast to float32 online during training to enable augmentation functions that expect float32 to still work

## Why
* Allows for larger training datasets to be used without exceeding memory

Partially addresses #382. Addresses all of the memory concerns in #403.


deepcell/image_generators/__init__.py
```diff
@@ -173,9 +174,9 @@ def _transform_masks(y, transform, data_format=None, **kwargs):
         beta = kwargs.pop('beta', 1)
 
         if data_format == 'channels_first':
-            y_transform = np.zeros(tuple([y.shape[0]] + list(y.shape[2:])))
+            y_transform = np.zeros(tuple([y.shape[0]] + list(y.shape[2:])), dtype='float16')
         else:
-            y_transform = np.zeros(y.shape[0:-1])
+            y_transform = np.zeros(y.shape[0:-1], dtype='float16')
```
willgraf, line 179: why are these float16, but later we re-cast to uint8?

Should we have an optional kwarg for the target dtype? It can default to uint8 but will allow a bit more customization.

deepcell/image_generators/semantic.py
```diff
@@ -105,8 +105,16 @@ def __init__(self,
...
+        if X.dtype in ['float32', 'int32', 'float64', 'int64']:
+            logging.warning('X data dtype is {}: this will increase memory use during '
+                            'preprocessing. Consider using a smaller dtype'.format(X.dtype))
+
+        if y.dtype in ['float32', 'int32', 'float64', 'int64']:
+            logging.warning('y data dtype is {}: this will increase memory use during '
+                            'preprocessing. Consider using a smaller dtype.'.format(y.dtype))
+
+        self.x = np.asarray(X, dtype='float16')
+        self.y = np.asarray(y, dtype='int16')
```
willgraf, line 117: It looks like we warn the user and then throw away the dtype we warn them about. Maybe we should update the warning message to tell them what dtype it will be cast to?

deepcell/image_generators/__init__.py
ngreenwald: I wanted to pick the smallest dtype appropriate for that location. I used float16 since the inner distance transform gives floats, but then if it get binned and changed to one-hot, it can be converted to ints

deepcell/image_generators/semantic.py
ngreenwald: It turns out that the X and y that are passed in also take up significant memory. So either way, we convert to float16, but that's not enough to get rid fo the memory cost of loading them in the first place if there were float64, for example

ngreenwald:
I don't see a downside to having the image_generators take an argument for default precision or something like that.

ngreenwald:
Added a test to make sure dtypes were set as expected, and also added functionality to croppinggenerator

willgraf (changes requested):
Coming back to this PR...I wonder if I was wrong about having the default values in the image data generator. If we just had an optional `dtype` parameter in `transforms_kwargs` it seems like it would be a minimal change to support this functionality.

ngreenwald:
I couldn't find a way to pass the default dtype of y, since I wanted to avoid having a separate int_dtype and float_dtype. Is there some way to figure out, based on y.dtype, what the bit size is?

deepcell/image_generators/semantic.py
```diff
@@ -367,10 +375,11 @@ def random_transform(self, x, y=None, seed=None):
         """
         params = self.get_random_transform(x.shape, seed)
 
+        # temporarily convert to float32 so enable scipy.nd_image functions to work
```
willgraf, line 378: We say temporarily, but do we cast it back?

deepcell/image_generators/semantic.py
```diff
@@ -117,14 +122,15 @@ def __init__(self,
         self.save_format = save_format
         self.min_objects = min_objects
 
+        self.x = np.asarray(X)
+        self.y = np.asarray(y)
+
```
willgraf, line 127: What is this for? If they're not already arrays, wouldn't it crash on the dtype check above

deepcell/image_generators/semantic.py
```diff
@@ -106,8 +106,13 @@ def __init__(self,
...
 
-        self.x = np.asarray(X, dtype=K.floatx())
-        self.y = np.asarray(y, dtype='int32')
+        if X.dtype in ['float32', 'int32', 'float64', 'int64']:
+            logging.warning('X data dtype is {}: this will increase memory use during '
+                            'preprocessing. Consider using a smaller dtype'.format(X.dtype))
+
+        if y.dtype in ['float32', 'int32', 'float64', 'int64']:
+            logging.warning('y data dtype is {}: this will increase memory use during '
+                            'preprocessing. Consider using a smaller dtype.'.format(y.dtype))
```
willgraf, line 115: Should we warn on 32? I think that's the standard size. With the other batching PR we will have far fewer memory restrictions too, so 32 might not be an obstacle.

deepcell/image_generators/semantic.py
```diff
@@ -135,11 +141,13 @@ def __init__(self,
...
             for transform in transforms:
                 transform_kwargs = transforms_kwargs.get(transform, dict())
+                # add default dtype to kwargs
+
                 y_transform = _transform_masks(y_current, transform,
                                                data_format=data_format,
                                                **transform_kwargs)
                 if y_transform.shape[self.channel_axis] > 1:
-                    y_transform = np.asarray(y_transform, dtype='int32')
+                    y_transform = np.asarray(y_transform, dtype=self.y.dtype)
```
willgraf, line 150: Was there an issue where multi-channel y were getting forced to ints when they shouldn't?

Do we need this conditional anymore? Or should we just always use `self.y.dtype` regardless of the channel size of y?

deepcell/image_generators/semantic.py
```diff
@@ -135,11 +141,13 @@ def __init__(self,
 
             for transform in transforms:
                 transform_kwargs = transforms_kwargs.get(transform, dict())
+                # add default dtype to kwargs
+
```
willgraf, line 145: I don't think we're doing this anymore.

```suggestion
```

willgraf (changes requested):
I left a few comments/questions.

I'm wondering if we should merge this into the transform-per-batch PR (or vice versa) before going into master? They both are attempting to solve the same problem - high memory usage in the data generators.

deepcell/image_generators/semantic.py
ngreenwald: I have no idea. Initially I just deleted it (it was on line 109 before), but I got some pretty low-level error messages about index not being a numpy array when I did. So somewhere, something is generating a list.

deepcell/image_generators/semantic.py
ngreenwald: No, I guess it's not actually temporary, just on a per-batch basis which has a small memory footprint

deepcell/image_generators/semantic.py
ngreenwald: I think that issue was the bug in the HPS branch for doing online transforms, not 100% sure. I don't think I ran into anything similar.

We can't use self.y.dtype, because if the dtype is `int`, and we have inner-distance transform which produces floats from 0-1, it'll ruin the output. 

Maybe we could always use float? I vaguely remember one of these transforms expending ints for 0/1 equivalent of boolean, but I'm not sure

deepcell/image_generators/semantic.py
ngreenwald: Agreed, makes sense

willgraf:
Closing this PR in favor of #436

PR #413, opened 2020-07-04
Updated multiplex app and new postprocessing function

### LINKED ISSUE #411: Weights for MultiplexApplication redownload
**Describe the bug**
Weights for the multiplex application are redownloaded due to an incorrect hash. This may be due to the hash value being incorrect.

**Expected behavior**
Model weights should download once, and not be re-downloaded if the application object is re-instantiated.

**Desktop (please complete the following information):**
Ubuntu 16.04

ngreenwald:
## What
* Updates the multiplex_segmentation app to support models that predict multiple distinct cellular compartments. 

## Why
* To enable predicting both nuclear and whole-cell signal

## TODOs:
I put the new postprocessing function in here just to make it easier to review both together. Once you're both happy with the approach and implementation I'll migrate to a separate PR in deepcell-toolbox.

I changed some of the kw arguments in the postprocessing function because I think the ones we're currently using aren't that intuitive. However, if you think that is going to cause confusion I can switch them back to match the existing defaults. 

- [x] Update the weights file with the latest and greatest model once this is ready to be merged in
- [x] Update the hash. This closes #411

### DIFF
--- a/deepcell/applications/application.py
+++ b/deepcell/applications/application.py
@@ -54,6 +54,7 @@ class Application(object):
         Raises:
             ValueError: `Preprocessing_fn` must be a callable function
             ValueError: `Postprocessing_fn` must be a callable function
+            ValueError: `Model_output_fn` must be a callable function
         """
 
     def __init__(self,
@@ -62,6 +63,7 @@ class Application(object):
                  model_mpp=0.65,
                  preprocessing_fn=None,
                  postprocessing_fn=None,
+                 format_model_output_fn=None,
                  dataset_metadata=None,
                  model_metadata=None):
 
@@ -76,6 +78,7 @@ class Application(object):
         self.model_mpp = model_mpp
         self.preprocessing_fn = preprocessing_fn
         self.postprocessing_fn = postprocessing_fn
+        self.format_model_output_fn = format_model_output_fn
         self.dataset_metadata = dataset_metadata
         self.model_metadata = model_metadata
 
@@ -84,6 +87,8 @@ class Application(object):
             raise ValueError('Preprocessing_fn must be a callable function.')
         if self.postprocessing_fn is not None and not callable(self.postprocessing_fn):
             raise ValueError('Postprocessing_fn must be a callable function.')
+        if self.format_model_output_fn is not None and not callable(self.format_model_output_fn):
+            raise ValueError('Format_model_output_fn must be a callable function.')
 
     def predict(self, x):
         raise NotImplementedError
@@ -216,6 +221,24 @@ class Application(object):
 
         return output_images
 
+    def _format_model_output(self, output_images):
+        """Applies formatting function the output from the model if one was provided.
+        Otherwise, returns the model output unmodified
+
+        Args:
+            output_images: stack of untiled images to be reformatted
+
+        Returns:
+            dict or list: reformatted images stored as a dict, or input images stored as list
+                if no formatting function is specified
+        """
+
+        if self.format_model_output_fn is not None:
+            formatted_images = self.format_model_output_fn(output_images)
+            return formatted_images
+        else:
+            return output_images
+
     def _resize_output(self, image, original_shape):
         """Rescales input if the shape does not match the original shape
         excluding the batch and channel dimensions
@@ -273,7 +296,10 @@ class Application(object):
         # Untile images
         output_images = self._untile_output(output_tiles, tiles_info)
 
-        return output_images
+        # restructure outputs into a dict if function provided
+        formatted_images = self._format_model_output(output_images)
+
+        return formatted_images
 
     def _predict_segmentation(self,
                               image,
--- a/deepcell/applications/application.py
+++ b/deepcell/applications/application.py
@@ -50,6 +50,8 @@ class Application(object):
             postprocessing_fn (function, optional): Postprocessing function to apply
                 to data after prediction. Defaults to None.
                 Must accept an input of a list of arrays and then return a single array.
+            format_model_output_fn (function, optional): Convert model output from a list
+                of matrices to a dictionary with keys for each semantic head
 
         Raises:
             ValueError: `Preprocessing_fn` must be a callable function
--- a/deepcell/applications/multiplex_segmentation.py
+++ b/deepcell/applications/multiplex_segmentation.py
@@ -130,7 +130,7 @@ class MultiplexSegmentation(Application):
                 os.path.basename(WEIGHTS_PATH),
                 WEIGHTS_PATH,
                 cache_subdir='models',
-                md5_hash='66fec859eacc5222b5e7d2baa105f3e3'
+                md5_hash='a2f26a7b5cc9a86d68e65a7bd27c32d0'
             )
 
             model.load_weights(weights_path)
@@ -170,7 +170,7 @@ class MultiplexSegmentation(Application):
                 [whole-cell, nuclear, both]
             postprocess_kwargs_whole_cell (dict, optional): Kwargs to pass to postprocessing
                 function for whole_cell prediction. Defaults to {}.
-           postprocess_kwargs_nuclear (dict, optional): Kwargs to pass to postprocessing
+            postprocess_kwargs_nuclear (dict, optional): Kwargs to pass to postprocessing
                 function for nuclear prediction. Defaults to {}.
 
         Raises:
--- a/deepcell/applications/multiplex_segmentation.py
+++ b/deepcell/applications/multiplex_segmentation.py
@@ -39,8 +39,6 @@ from deepcell_toolbox.processing import phase_preprocess
 from deepcell.applications import Application
 from deepcell.model_zoo import PanopticNet
 
-import numpy as np
-
 
 WEIGHTS_PATH = ('https://deepcell-data.s3-us-west-1.amazonaws.com/'
                 'model-weights/Multiplex_Segmentation_20200714_subcellular.h5')
--- a/deepcell/applications/multiplex_segmentation.py
+++ b/deepcell/applications/multiplex_segmentation.py
@@ -203,11 +203,16 @@ class MultiplexSegmentation(Application):
                  use_pretrained_weights=True,
                  model_image_shape=(256, 256, 2)):
 
+        whole_cell_classes = [1, 1, 2, 3]
+        nuclear_classes = [1, 1, 2, 3]
+        num_semantic_classes = whole_cell_classes + nuclear_classes
+        num_semantic_heads = len(num_semantic_classes)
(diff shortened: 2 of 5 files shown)


deepcell/applications/multiplex_segmentation.py
```diff
@@ -40,6 +40,75 @@
...
+import numpy as np
+
+
+def deep_watershed_subcellular(outputs, compartment='cell', min_distance=10, maxima_threshold=0.1,
+                               cell_threshold=0.1, exclude_border=False,
+                               small_objects_threshold=0):
+    """Postprocess model output to generate predictions for distinct cellular compartments
+
+    Args:
+        outputs (list): DeepWatershed model output. A list of
```
msschwartz21, line 52: Can we assume that the order of the compartments in the list is always the same? If so, let's make a note of the appropriate order in the docs

deepcell/applications/multiplex_segmentation.py
```diff
@@ -40,6 +40,75 @@
...
+    """
+
+    valid_compartments = ['cell', 'nucleus', 'both']
+
+    if compartment not in valid_compartments:
+        raise ValueError('Invalid compartment supplied: {}. '
+                         'Must be one of {}'.format(compartment, valid_compartments))
+
+    if compartment == 'cell':
+        label_images = deep_watershed_mibi(outputs=outputs[:4],
```
msschwartz21, line 80: It seems like you are assuming that outputs will always be a list of length 8 first listing cell data and then nuclear data. Is that a correct/safe assumption?

deepcell/applications/multiplex_segmentation.py
```diff
@@ -40,6 +40,75 @@
...
+        compartment: which cellular compartments to generate predictions for.
+            must be one of 'cell', 'nucleus', 'both'
+        min_distance (int): Minimum allowable distance between two maxima to seed cells.
+        maxima_threshold (float): Threshold for maxima used to define watershed seeds.
+        cell_threshold (float): Threshold for mask used to define allowable area for watershed.
+        exclude_border (bool): Whether to include centroid detections at the border.
+        small_objects_threshold (int): Removes objects smaller than this size.
+
+    Returns:
+        numpy.array: Uniquely labeled mask for each compartment
```
msschwartz21, line 67: Can we explicitly state the order of the masks, e.g. nuclear then cell and if possible which axis is the "channel" axis?

ngreenwald:
Hi Morgan, 
Those assumptions are correct for the current version of the model. Because it was trained with the cell labels passed to the first 4 semantic heads and the nuclear labels passed to the second 4 semantic heads, we'll always have (for this model) a list of 8 with the first four being cell and the second four being nuclear. 

However, the longer-term goal is instantiate the this model only with the heads necessary for prediction, which would change this to a list of 4 with the first two being cell and the second nuclear. 

So I imagine that this could end up changing. Given that, I'm open to what you and @willgraf think the best way to handle that would be.

msschwartz21:
My initial thought was to ask if we could change the data structure of `outputs` to be a dictionary where the keys are 'cell' or 'nucleus' and then the values are the prediction lists? Then the processing function could just loop over each item in the dictionary since it doesn't seem like the processing function call is different. But after a second look at the code, it doesn't seem like there is a convenient place to change the output data structure. Thoughts?

ngreenwald:
I like that idea! What about adding an Application-specific labeling step that takes the output of the model and puts it in such a datastructure?

It could go in _run_model after the untiled list of images gets returned:
https://github.com/vanvalenlab/deepcell-tf/blob/e6d6d6b18eca5bbf2c5c44a0b3d695df4fd54470/deepcell/applications/application.py#L274

msschwartz21:
That seems like a logical place for it to fit in. Were you thinking about overwriting `_run_model` in the multiplex application or permanently adding a conditional flag to the base function?

ngreenwald:
I was thinking of modeling it off of the preprocessing function, which gets
run if there is one supplied, across the different applications. Does
that make sense?

deepcell/applications/multiplex_segmentation.py
willgraf: I think the `Application` classes automatically assume the channel axis is `-1`, unless I'm misunderstanding.

deepcell/applications/multiplex_segmentation.py
willgraf: Yeah maybe a more flexible solution is to assume that there are 2 sets of results, and instead of `outputs[:4]` we can use something like `outputs[:len(outputs)//2]`.

willgraf:
Hey sorry, a little late to the party.

> However, the longer-term goal is instantiate this model only with the heads necessary for prediction, which would change this to a list of 4 with the first two being cell and the second nuclear.

This is pretty straightforward, we just need to load the weights with `by_name=True` to drop the unused inference heads. Any reason not to shave off the unused heads in this PR?

> I was thinking of modeling it off of the preprocessing function, which gets
run if there is one supplied, across the different applications. Does
that make sense?

I think this is a good solution, if I'm following the conversation.

ngreenwald:
Oh, that's super easy. Alright, I'll get started on that

deepcell/applications/application.py
```diff
@@ -62,6 +63,7 @@ def __init__(self,
                  model_mpp=0.65,
                  preprocessing_fn=None,
                  postprocessing_fn=None,
+                 format_model_output_fn=None,
```
willgraf, line 66: Can we document what this does in the docstring?

deepcell/applications/multiplex_segmentation.py
```diff
@@ -40,6 +40,100 @@
...
+                                           distance_threshold=cell_threshold,
+                                           exclude_border=exclude_border,
+                                           small_objects_threshold=small_objects_threshold)
+    elif compartment == 'nuclear':
+        label_images = deep_watershed_mibi(outputs=model_output['nuclear'],
+                                           min_distance=min_distance,
+                                           detection_threshold=maxima_threshold,
+                                           distance_threshold=cell_threshold,
+                                           exclude_border=exclude_border,
+                                           small_objects_threshold=small_objects_threshold)
```
willgraf, line 116: These two look almost the same.

```suggestion
    if compartment in {'whole_cell', 'nuclear'}:
        label_images = deep_watershed_mibi(outputs=model_output[compartment],
                                           min_distance=min_distance,
                                           detection_threshold=maxima_threshold,
                                           distance_threshold=cell_threshold,
                                           exclude_border=exclude_border,
                                           small_objects_threshold=small_objects_threshold)
```

deepcell/applications/multiplex_segmentation.py
```diff
@@ -40,6 +40,100 @@
...
+        small_objects_threshold (int): Removes objects smaller than this size.
+
+    Returns:
+        numpy.array: Uniquely labeled mask for each compartment
+
+    Raises:
+        ValueError: for invalid compartment flag
+    """
+
+    valid_compartments = ['whole_cell', 'nuclear', 'both']
```
willgraf, line 97: Generally better to ask if something is in a `set` rather than in a `list` (O(1) vs O(3)).

```suggestion
    valid_compartments = {'whole_cell', 'nuclear', 'both'}
```

deepcell/applications/multiplex_segmentation.py
```diff
@@ -112,8 +206,8 @@ def __init__(self,
         model = PanopticNet('resnet50',
                             input_shape=model_image_shape,
                             norm_method=None,
-                            num_semantic_heads=4,
-                            num_semantic_classes=[1, 1, 2, 3],
+                            num_semantic_heads=8,
+                            num_semantic_classes=[1, 1, 2, 3, 1, 1, 2, 3],
```
willgraf, line 210: I do not like the list of 8 numbers. Maybe we can alias them?

```python
whole_cell_classes = [1, 1, 2, 3]
nuclear_classes = [1, 1, 2, 3]

num_semantic_classes=whole_cell_classes + nuclear_classes
num_semantic_heads=len(num_semantic_heads)  # this is always true and should not be a required parameter.

**Comment by ngreenwald** [comment]
Great suggestions, thanks Will.

**Comment by ngreenwald** [comment]
Yeah, I'm going to need to change deepcell_toolbox.deep_watershed_mibi to handle the newly formatted data. Just wanted to make sure it all looked good before I start changing stuff over there.

**Comment by ngreenwald** [comment]
I guess that's another thing to consider. If we're going to have all of the model accessory/processing functions in the toolbox, then we'll have to patch the toolbox when we update them. Not a huge deal, and there's no rush on getting this PR integrated into master.

**Comment by ngreenwald** [comment]
I had to be a bit creative with fitting the new post-processing dictionaries into the existing post-processing kwargs argument in the main application. Let me know if you think it's smart or hacky

File: deepcell/applications/multiplex_segmentation.py [hunk 8]
```diff
@@ -156,8 +166,12 @@ def predict(self,
...
             preprocess_kwargs (dict, optional): Kwargs to pass to preprocessing function.
                 Defaults to {}.
-            postprocess_kwargs (dict, optional): Kwargs to pass to postprocessing function.
-                Defaults to {}.
+            compartment (string): Specify type of segmentation to predict. Must be one of
+                [whole-cell, nuclear, both]
+            postprocess_kwargs_whole_cell (dict, optional): Kwargs to pass to postprocessing
+                function for whole_cell prediction. Defaults to {}.
+           postprocess_kwargs_nuclear (dict, optional): Kwargs to pass to postprocessing
+                function for nuclear prediction. Defaults to {}.
```
willgraf, line 174: spacing looks a little off here.

deepcell/applications/multiplex_segmentation.py
```diff
@@ -169,6 +183,12 @@ def predict(self,
             np.array: Labeled image
             np.array: Model output
         """
+
+        # create dict to hold all of the post-processing kwargs
+        postprocess_kwargs = {'whole_cell_kwargs': postprocess_kwargs_whole_cell,
+                              'nuclear_kwargs': postprocess_kwargs_nuclear,
+                              'compartment': compartment}
```
willgraf, line 190: so `compartment` could be `nuclear`, `whole_cell` or `both`. If it is `both` then both kwargs are needed, and if it is one of the others, then only one kwargs is needed?

The post-processing will catch any bad values, correct?

ngreenwald: The kwargs aren't needed, since they default to be empty and all of the parameters in the actual function have default values which are only overriden in the case that they are supplied here. But yes, to override a value in nuclear predictions the nuclear_kwargs dict has to be provided.

PR #416, opened 2020-07-20
3 d nuclear segmentation

cdpavelchek:
## What
Fixed bugs in 3D model training

## Why
3D model training was broken and needed to be fixed.


deepcell/model_zoo/panopticnet.py
```diff
@@ -76,26 +76,35 @@ def __merge_temporal_features(feature, mode='conv', feature_size=256,
             raise ValueError('Mode {} not supported. Please choose '
                              'from {}.'.format(mode, str(acceptable_modes)))
 
+    f_name = str(feature.name)[:2]
+
+    conv_name = 'conv_mtf_{}'.format(f_name)
+    bnorm_name = 'bnorm_mtf_{}'.format(f_name)
+    acti_name = 'acti_mtf_{}'.format(f_name)
```
willgraf, line 83: If these are only used inside one of the blocks, just define them in that block (maybe even just skip the var name and pass in the string to the layer directly).

deepcell/utils/backbone_utils.py
```diff
@@ -465,8 +465,10 @@ class Utils(object):  # pylint: disable=useless-object-inheritance
 
         time_distributed_outputs = []
         for i, out in enumerate(layer_outputs):
+            td_name = 'td{}'.format(i)
+            md_name = 'md{}'.format(i)
```
willgraf, line 469: I like `td` shorthand, but I'm not sure about `md`. I don't think there's any harm in having a more descriptive name.

deepcell/utils/io_utils.py
```diff
@@ -159,8 +164,12 @@ def get_images_from_directory(data_location, channel_names):
         for j in range(n_channels):
             img_path = os.path.join(data_location, img_list_channels[j][stack_iteration])
             channel_img = get_image(img_path)
-            if data_format == 'channels_first':
+            if (data_format == 'channels_first') and (channel_img.ndim > 2):
+                all_channels[0, j, :, :, :] = channel_img
+            elif data_format == 'channels_first':
                 all_channels[0, j, :, :] = channel_img
+            elif channel_img.ndim > 2:
+                all_channels[0, :, :, :, j] = channel_img
```
willgraf, line 172: I think this can be simplified by not enumerating each dimension:

```python
all_channels[0, j, ...] = channel_img

# or

all_channels[0, ..., j] = channel_img
```

deepcell/utils/plot_utils.py
```diff
@@ -59,16 +59,19 @@ def get_js_video(images, batch=0, channel=0, cmap='jet', vmin=0, vmax=30):
         'cmap': cmap,
     }
 
+    if vmax == 0:
+        vmax = images.max()
+
     # TODO: do these not work for other cmaps?
-    if cmap == 'cubehelix':
+    if cmap == 'cubehelix' or cmap == 'jet':
```
willgraf, line 66: Do you know what's up with the color maps? Should these be enumerated, or just check if `cmap is None`?

deepcell/utils/plot_utils.py
```diff
@@ -59,16 +59,19 @@ def get_js_video(images, batch=0, channel=0, cmap='jet', vmin=0, vmax=30):
         'cmap': cmap,
     }
 
+    if vmax == 0:
+        vmax = images.max()
```
willgraf, line 63: Why did this work before? Should we instead default `vmax=None`?

deepcell/utils/plot_utils.py
```diff
@@ -59,16 +59,19 @@ def get_js_video(images, batch=0, channel=0, cmap='jet', vmin=0, vmax=30):
...
+    if cmap == 'cubehelix' or cmap == 'jet':
         plot_kwargs['vmin'] = vmin
         plot_kwargs['vmax'] = vmax
 
     for i in range(images.shape[1]):
         im = plt.imshow(images[batch, i, :, :, channel], **plot_kwargs)
         ims.append([im])
 
-    ani = animation.ArtistAnimation(fig, ims, interval=150, repeat_delay=1000)
+    ani = animation.ArtistAnimation(fig, ims, interval=interval, repeat_delay=1000)
```
willgraf, line 74: Should we add `repeat_delay` while we're at it?

deepcell/utils/io_utils.py
```diff
@@ -149,8 +149,13 @@ def get_images_from_directory(data_location, channel_names):
...
 
     for stack_iteration in range(len(img_list_channels[0])):
-        if data_format == 'channels_first':
+
+        if (data_format == 'channels_first') and (img_temp.ndim > 2):
+            shape = (1, n_channels, img_temp.shape[0], img_temp.shape[1], img_temp.shape[2])
+        elif data_format == 'channels_first':
             shape = (1, n_channels, img_temp.shape[0], img_temp.shape[1])
+        elif img_temp.ndim > 2:
+            shape = (1, img_temp.shape[0], img_temp.shape[1], img_temp.shape[2], n_channels)
```
willgraf, line 158: Is this just for loading 3d data? what was the issue here?

deepcell/model_zoo/panopticnet.py
cdpavelchek: Done

deepcell/utils/backbone_utils.py
cdpavelchek: done

deepcell/utils/io_utils.py
cdpavelchek: yeah, done

deepcell/utils/plot_utils.py
cdpavelchek: I'm not sure what the TODO is referencing. Maybe there are certain cmaps it doesn't work with and it was safer to do this than try and list all the exceptions? Not really sure, and haven't needed any other cmaps so far

deepcell/utils/plot_utils.py
cdpavelchek: It defaulted to a max of 30 - for some reason the docs had it as the vmax being number of frames? We could default it to None, doesn't really matter.

deepcell/utils/plot_utils.py
cdpavelchek: Might as well, done

deepcell/utils/io_utils.py
cdpavelchek: Yeah it's for loading 3D data, would break if the image didn't have ndim==2

willgraf: I think we can clean up the `if` blocks by not enumerating each shape but just concatenating lists.

```python
if data_format == 'channels_first':
    shape = tuple([1, n_channels] + list(img_temp.shape))
else:
    shape = tuple([1] + list(img_temp.shape) + [n_channels])
```

cdpavelchek: Done

PR #429, opened 2020-09-11
Move wrapper functions to applications

ngreenwald:
## What
- Moves the simple wrapper functions which are application-specific to the applications file.
- Updates requirements.txt for new version of the toolbox, which this PR relies on

## Why
- These functions aren't broadly useful, and don't really belong in the toolbox. Moving them here also makes it easier to update things and doesn't require patches across repos. This will also be useful once this whole module gets moved to its own repo

## Remaining tasks
- Update weights file once new model is trained
- Update post-processing default parameters once HPS finishes


ngreenwald:
Added in new weights files, as well as the current settings I'm using for postprocessing. 
Updated the model and test functions to expect a 2 semantic heads per prediction instead of 4.

willgraf (changes requested):
Looking at the redis-consumer PR makes me realize that we probably made a mistake and should keep the format_multiplex and multiplex wrappers in the toolbox - specifically so that they are able to be used by the applications AND the consumer. I did not realize how much they are used by the consumers.

I'll try to set up a PR that moves them back into the toolbox and we'll release a 0.7.1 (if that makes sense to you).

requirements-test.txt
```diff
@@ -2,4 +2,5 @@
 pytest==4.6.5
 pytest-cov==2.5.1
 pytest-pep8>=1.0.6,<2.0.0
+pytest-mock
```
willgraf, line 5:
```suggestion
```

deepcell/applications/multiplex_segmentation_test.py
```diff
@@ -29,6 +29,7 @@
 from __future__ import division
 from __future__ import print_function
 
+
```
willgraf, line 32:
```suggestion
```

willgraf (changes requested):

willgraf (approved):

PR #436, opened 2020-09-25
Reduce ImageDataGenerator memory footprint.

### LINKED ISSUE #382: Out of memory errors - change data generators

willgraf:
## What
* Change the `SemanticIterators` from pre-computing all the transforms to computing the transform of every batch inside `_get_batches_of_transformed_samples`.
* Preserve `dtype` when creating batches of samples in all generators.
* Use an appropriate `dtype` when calling `_transform_masks` so that we are not using `int64` arrays when there are only 2 unique values.

## Why
* This change should significantly reduce the memory footprint of the `SemanticIterators`. Instead of saving the full set of data for every transform (up to 7x the data in some cases), the data is computed on the fly and takes a minimal amount of extra space. While this does increase the training time per epoch, it significantly reduces the resource requirements of the training environment.
* Closes #382


willgraf:
I've taken some valuable changes from #412 and tried to implement them here in this PR to further reduce the memory footprint of each data generator.

deepcell/image_generators/image_generators_test.py
```diff
@@ -66,22 +66,26 @@ def test_no_transform(self):
         mask_transform = image_generators._transform_masks(
             mask, transform=None, data_format='channels_last')
         self.assertEqual(mask_transform.shape, (5, 30, 30, num_classes))
+        self.assertTrue(np.issubdtype(mask_transform.dtype, np.integer))
```
ngreenwald, line 69: Not every test here needs it, but it would be good to at least spot-check that the correct dtype is being used. e.g. if we assign K.floatx() to be 'float16', that the transforms are using it. I think all of these tests would pass if no downsizing was actually being performed.

ngreenwald (changes requested):
Looks good, overall super clean. 

The reason we had to cast to float32 in the previous PR for `apply_transform` was not due to issues with half-precision computation, but because the open_cv transforms that are used as part of augmentation fail with float16 images. Will this new version run without failing if K.floatx() is set to 'float16'? I think this could be another reason to actually test out different values of K.floatx() as I mentioned in my comment below.

deepcell/image_generators/image_generators_test.py
willgraf: Good call, I added it to both the distance transforms tests since I think they are the only float outputs.

willgraf:
> The reason we had to cast to float32 in the previous PR for apply_transform was not due to issues with half-precision computation, but because the open_cv transforms that are used as part of augmentation fail with float16 images.

Thank you, I see now that you commented it fails during the [affine transforms](https://github.com/keras-team/keras-preprocessing/blob/5949df1c059a53d98a6004d5bfc93708e5ec6c4a/keras_preprocessing/image/affine_transformations.py#L331) due to scipy.ndimage.

I'm not sure the best way to handle it. If the user explicitly sets floatx to float16, should we re-cast to float32 in `random_transform`? Maybe just let them use float16 and warn the user? I'll dig a bit deeper into which transforms are not half-precision friendly.

ngreenwald:
Yeah, I think casting and giving a warning seems fine. From what I remember, there wasn't an issue that would lead to inaccurate results, only came up because of some checks that were automatically performed by keras for dtypes.

willgraf:
A little more digging on float16 compatibility:

`scipy.ndimage.interpolation.affine_transform` is not compatible with `float16`, which means that none of the following options are compatible:

- `rotation_range`
- `zoom_range`
- `height_shift_range`
- `width_shift_range`
- `shear_range`

To make `apply_transform` compatible with `float16`, we would need to update each data generator with the float32-casting.

In addition, `float16` is not currently supported by the `PanopticNet` models.

willgraf:
Considering all the incompatibilities with `float16`, I think we don't need to update `apply_transform` to account for the user setting `K.set_floatx('float16')`.

ngreenwald:
I agree, the only reason I was using float16 was to further reduce memory footprint when everything was pre-computed. I don't see any scenario for a single batch of images where float16 vs float32 makes a noticeable difference in memory

PR #446, opened 2020-10-30
Add K.epsilon to denominator in whole_image normalization.

### LINKED ISSUE #445: whole-image normalization has no epsilon in the denominator
**Describe the bug**
The preprocessing functions image normalization, which z-score each channel in each image, don't have an epsilon in the denominator of this calculation. For images with very low standard deviation, this causes the values to blow up. 
This is true both for the version in [deepcell-tf](https://github.com/vanvalenlab/deepcell-tf/blob/83226abe7a7ca5372d42be3994963d1fa19c0b34/deepcell/layers/normalization.py#L158) and the [toolbox](https://github.com/vanvalenlab/deepcell-toolbox/blob/07cdda1195b0630765c2c2ac41c28c1ed0ec71bb/deepcell_toolbox/processing.py#L43)

This caused weird bugs during training, which were resolved if I modified the above function to have a constant factor of 0.001 in the denominator of the z-score calculation

willgraf:
## What
* Add `K.epslion()` to the denomiator of the whole_image normalization calculation.
* Update `deepcell-toolbox` to 0.8.1.

## Why
* Prevent any `NaN` errors if the standard deviation is 0.
* Fixes #445


willgraf:
Before merging this in I would like to update the version of `deepcell_toolbox` to the version with the `normalize` patch.

ngreenwald (changes requested):
Sure thing, makes sense

PR #458, opened 2020-11-12
Refresh all notebooks with TensorFlow 2 - remove training.py calls.

willgraf:
## What
* Refresh all the notebooks with the new TensorFlow 2 library.
* Remove all `train_model` functions and replace them with the contents of the script.

## Why
* Make sure all notebooks run well with TensorFlow 2.
* Make notebooks more clear by directly training models rather than calling a trainer function.


willgraf:
The `disc` notebooks do not run well.
 
- 2D disc loss has an issue during post-processing (MeanShift and DBSCAN take a VERY long time)
- 3D has trouble during training the `disc` model. I think it may be related to the loss function, though I'm not positive.

PR #460, opened 2020-11-17
Update Applications to use SavedModel format

willgraf:
## What
* Replace the hard-coded model architecture and replace it with loading a model tar file from S3.
* Add `LabelDetection` and `ScaleDetection` applications.

## Why
* This makes the models themselves much more robust to small changes in the codebase. For example, if an architecture removes a layer, the application will not need to be retrained to be usable, as the saved model will still contain the removed layer.


willgraf:
The `CellTracking` and `CytoplasmSegmentation` applications have not been updated yet. To do so requires model retraining, which I am not prepared to do at the moment. The current old-style applications will continue to work, though the results will be of dubious quality since the weights were trained with an older version of TensorFlow.

PR #470, opened 2021-01-08
Migrate DeepCell Tracking to TF2

MekWarrior:
## What
* Change the tracking data generator and model to support batches as dictionaries rather than one array. Adapt the train model function to use tf.datasets based on the new data generator.

## Why
* The old tracking model, data generator, and train model function did not support TF2. These adaptations complete deepcell-tf's migration to the latest version of TensorFlow.


deepcell/training.py
```diff
@@ -515,6 +513,61 @@ def loss_function(y_true, y_pred):
...
+        input_type_dict['{}_input1'.format(feature)] = tf.float32
+        input_type_dict['{}_input2'.format(feature)] = tf.float32
+
+        if feature == 'appearance':
+            app1 = tuple([None, train_data.min_track_length,
+                          train_data.crop_dim, train_data.crop_dim, 1])
+            app2 = tuple([None, 1, train_data.crop_dim, train_data.crop_dim, 1])
+
+            input_shape_dict['{}_input1'.format(feature)] = app1
+            input_shape_dict['{}_input2'.format(feature)] = app2
```
willgraf, line 533: The formatting of these names happens a lot in these blocks. Could we just define them once at the top and reuse it? I think it could really clean up some of this if/else block.

willgraf (commented):
The generator changes look good, but I'm hopeful we can clean up the training.py script a bit before merging.

deepcell/training.py
MekWarrior: Agreed. Moved as requested.

willgraf (approved):
I'm approving despite the reduced code coverage because:
- There is an [existing TODO](https://github.com/vanvalenlab/deepcell-tf/blob/a9afa683879b00422b7cdb0a841f202c4f02c0c7/deepcell/image_generators/image_generators_test.py#L1238) for correctness tests
- `training.py` is currently untested.

PR #473, opened 2021-01-23
Update Tracking Application

MekWarrior:
## What
* The `cell_tracking` application was updated from TF1-sytle with weights to TF2-style encapsulating the full model signature, respecting deepcell-tf pattern.

## Why
* The tracking model has been updated for TF2 and this PR updates DeepCell Applications to match.


deepcell/applications/cell_tracking_test.py
```diff
@@ -66,7 +66,7 @@ class TestCellTracking(test.TestCase):
     def test_cell_tracking_app(self):
         with self.cached_session():
             # test instantiation
-            app = CellTracking(use_pretrained_weights=False)
+            app = CellTracking()
```
willgraf, line 69: I suggest creating a dummy model here and passing it to `CellTracking`. That way the tests won't have to download anything from S3 and will prevent failures due to network errors or service outages.

willgraf (changes requested):
Looks good, just a small test change.

If possible could you post a gif of tracking results if you have anything handy?

deepcell/applications/cell_tracking_test.py
MekWarrior: Done.

willgraf (approved):

PR #497, opened 2021-02-26
Deprecate in_shape argument for Location layers.

willgraf:
## What
* Remove `in_shape` as an argument to `Location` layers; using the shape of inputs in `Layer.call` to determine the shape instead.
 
## Why
* `in_shape` is a non-standard layer input and should not be required to instantiate the layer. This change makes the `Location2D` and `Location3D` layers easier to work with and more flexible.


willgraf:
Re-ran the `PanOpticNet` notebook with `location=True` with expected results. Looks good!

PR #498, opened 2021-03-01
Use a dict of names and num_classes to name PanOpticNet semantic heads.

### LINKED ISSUE #414: Allow loading of any subset of semantic heads
Currently, to instantiate a model with only a subset of the semantic heads that were used during training, we use by_name=True. However, the semantic heads are named sequentially during model construction. This means whatever semantic head happens to be created first will be semantic_0, the second semantic head will be semantic_1, etc etc. 
https://github.com/vanvalenlab/deepcell-tf/blob/e6d6d6b18eca5bbf2c5c44a0b3d695df4fd54470/deepcell/model_zoo/panopticnet.py#L296

Therefore, if you try to remove a semantic head that wasn't last to be created, the other heads will get renumbered when the new model is created, leading to a weights mismatch during loading. 

Will suggested we change the way the layers associated with each semantic head are named. I can think of at least two options. The first would be to supply an additional argument to PanopticNet, which would be a list of names to use for the layers associated with each head. This is the simplest, but would result in an additional required argument. 

The second would be to list the type of transform you want each head to compute. For example, ['inner-distance', 'outer-distance', 'fgbg']. This list could then be used to the name the semantic head according the type of transform it is computing. We would then need to add a couple lines to get the appropriate number of semantic classes for that transform before this loop: 
https://github.com/vanvalenlab/deepcell-tf/blob/e6d6d6b18eca5bbf2c5c44a0b3d695df4fd54470/deepcell/model_zoo/panopticnet.py#L292
This would potentially replace the num_semantic_heads and num_semantic_classes lists. However, Dave mentioned that this would lead to much more rigidity during model construction, so might not be worth it. 

The current workaround for this is to have the semantic heads you want to have dropped during inference defined last in the model. However, this won't work for the models trained with multiple labels, because the semantic heads for each label get constructed in order.

willgraf:
## What
* Allow `num_semantic_heads` to be a dictionary instead of a list.
* Use dictionary keys as names of semantic heads instead of the index of the `num_semantic_heads` list.
* Convert `num_semantic_heads` from a list to a `dict` using the index as a key to maintain backward compatibility.
* Support transform names using either a hyphen or an underscore for better compatibility with layer names.

## Why
* Allows for more stable adding and removing of semantic heads.
* Fixes #414


willgraf:
Screen Shot 2021-03-01 at 11 39 27 AM

The output using the names to add/remove heads works! The F1 score is equivalent at ~0.8.

I think this is safe to say that the dictionary approach works as well.

PR #506, opened 2021-03-24
Add new GCNN based model architecture for tracking with new `tf.data.Dataset`

MekWarrior:
## What
* End-to-End updates for DeepCell's approach to the tracking problem in live-cell imaging. A dataset builder module was added to to more closely follow TensorFlow's preferred style (with tracking dataset object being the first example). The previous LSTM and Siamese Neural Network (SNN) architecture was moved from `featurenet.py` in the `model_zoo` to a new 'tracking.py,` where a graph-based architecture for classification was also installed. This new model required 2 new "merge" layers and an additional loss function compatible with masking out portions of the adjacency matrix. The Tracking Application has been updated to take advantage of this new model and approach.

## Why
* These updates represent the natural evolution of the repo's approach to tracking. It addresses key needs to further enable adoption of TF2 and dramatically improves tracking speed.


MekWarrior:
The tracking application test is correctly failing due a difference in `deepcell_tracking` that has not been released yet.

deepcell/model_zoo/README.md
```diff
@@ -1,9 +1,11 @@
...
 
-* [Siamese Model](https://www.biorxiv.org/content/10.1101/803205v2) This architecture, contained within featurenet.py, is used for cell tracking. It was inspired by [previous](http://openaccess.thecvf.com/content_iccv_2017/html/Sadeghian_Tracking_the_Untrackable_ICCV_2017_paper.html) object tracking models, and uses multiple features (cell appearance, neighborhood appearance, centroid motion, and morphological information) to link cells together over time.
+* [Siamese Model](https://www.biorxiv.org/content/10.1101/803205v2) This architecture is used for cell tracking. It was inspired by [previous](http://openaccess.thecvf.com/content_iccv_2017/html/Sadeghian_Tracking_the_Untrackable_ICCV_2017_paper.html) object tracking models, and uses multiple features (cell appearance, neighborhood appearance, centroid motion, and morphological information) to link cells together over time.
+
+* [Graph Tracking Model](https://www.biorxiv.org/content/10.1101/803205v2) This architecture uses a graph approach to accomplish cell tracking.
```
willgraf, line 11: This feels a little lacking compared to the other blurbs.

deepcell/model_zoo/tracking.py
```diff
@@ -0,0 +1,682 @@
...
+        self.unmerge_embeddings_model = self.get_unmerge_embeddings_model()
+        self.unmerge_centroids_model = self.get_unmerge_centroids_model()
+        self.embedding_temporal_merge_model = self.get_embedding_temporal_merge_model()
+        self.delta_temporal_merge_model = self.get_delta_temporal_merge_model()
+        self.appearance_encoder = self.get_appearance_encoder()
+        self.morphology_encoder = self.get_morphology_encoder()
+        self.centroid_encoder = self.get_centroid_encoder()
+        self.delta_encoder, self.delta_across_frames_encoder = self.get_delta_encoders()
+        self.neighborhood_encoder = self.get_neighborhood_encoder()
+        self.tracking_decoder = self.get_tracking_decoder()
```
willgraf, line 255: 💯

requirements.txt
```diff
@@ -3,6 +3,8 @@ scipy>=1.2.3,<2
 scikit-image>=0.14.5
 scikit-learn>=0.20.4
 tensorflow==2.4.1
+tensorflow_addons==0.12.1
+spektral==1.0.4
```
willgraf, line 7: These also need to be added to `setup.py`. They both exist due to the pip and docker builds.

deepcell/model_zoo/__init__.py
```diff
@@ -34,7 +34,8 @@
 from deepcell.model_zoo.featurenet import bn_feature_net_3D
 from deepcell.model_zoo.featurenet import bn_feature_net_skip_3D
 
-from deepcell.model_zoo.featurenet import siamese_model
+from deepcell.model_zoo.tracking import siamese_model
```
willgraf, line 37: Since you're changing the import path, make sure this is not disrupted in the existing tracking notebook.

deepcell/layers/__init__.py
```diff
@@ -43,6 +43,7 @@
 from deepcell.layers.normalization import ImageNormalization3D
 from deepcell.layers.pooling import DilatedMaxPool2D
 from deepcell.layers.pooling import DilatedMaxPool3D
+from deepcell.layers.temporal import TempMerge, TempUnmerge
```
willgraf, line 46: I personally think these should be renamed to `TemporalMerge` and `TemporalUnmerge` for more clarity.

deepcell/datasets/builders.py
```diff
@@ -0,0 +1,209 @@
...
+        self.track_length = track_length
+
+        # Create max_nodes and max_frames
+        self._create_max_sizes()
+
+        # Load tracks
+        self._load_tracks()
+
+        # Create datasets
+        self._create_dataset()
```
willgraf, line 60: Not a fan of the pattern, setting attributes in helper functions rather than in `__init__`.

deepcell/datasets/builders.py
```diff
@@ -0,0 +1,209 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+import tensorflow_addons as tfa
+
+
+class TrackingDatasetBuilder(object):
```
willgraf, line 37: Can this be [a subclass](https://www.tensorflow.org/datasets/api_docs/python/tfds/core/DatasetBuilder)?

deepcell/datasets/builders.py
```diff
@@ -0,0 +1,209 @@
...
+        y_dict = args[1]
+
+        app = X_dict['appearances']
+        max_time = tf.shape(app)[1] - self.track_length
+        t_start = tf.random.uniform([1],
+                                    minval=0,
+                                    maxval=max_time,
+                                    dtype=tf.int32)[0]
+
+        for key in X_dict.keys():
```
willgraf, line 149: dictionaries iterate over the keys by default, so there is no reason to call `.keys()` here, which creates a new list.

deepcell/datasets/builders.py
```diff
@@ -0,0 +1,209 @@
...
+        return (X_dict, y_dict)
+
+    def _create_dataset(self):
+        input_dict = {'appearances': self.appearances,
+                      'centroids': self.centroids,
+                      'morphologies': self.morphologies,
+                      'adj_matrices': self.norm_adj_matrices}
+        output_dict = {'temporal_adj_matrices': self.temporal_adj_matrices}
+        dataset = tf.data.Dataset.from_tensor_slices((input_dict, output_dict))
+        self.dataset = dataset.shuffle(256).repeat().map(self._sample_time).map(self._augment).batch(self.batch_size)
```
willgraf, line 205: It looks like all this class does is create this `dataset` value. I'm wondering if it even needs to be a class if that is the case.

willgraf (changes requested):
It would be nice to have some tests in `tracking_test.py`.

I would also like to see some test for `TemporalMerge` and `TemporalUnmerge`, but since these take in 2 inputs the standard `layer_test` routine won't work. You could check out the [RetinaMask layer tests](https://github.com/vanvalenlab/deepcell-retinamask/tree/master/deepcell_retinamask/layers) to see if there is a useful pattern there.

I'll need a walkthrough for both the dataset builder and the model as those are both pretty large.

deepcell/model_zoo/README.md
MekWarrior: My preference is to leave this blurb short until were closer to publication. We'll have more suitable text once we have an updated/new preprint.

requirements.txt
MekWarrior: Added

deepcell/model_zoo/__init__.py
MekWarrior: Both tracking notebooks (training and benchmarking will need to be updated for the new approach. Should be added before releasing the new preprint, but can be held until then.

deepcell/layers/__init__.py
MekWarrior: Names have been updated

deepcell/datasets/builders.py
MekWarrior: Corrected for clarity.

MekWarrior:
One outstanding blocker for this PR is the switch from transposed data. This will also need to be updated in `deepcell-tf`.

willgraf:
As discussed in #512 there are a few other TODOs after merging that PR into this branch.

- [The `deepcell.data.split_data` function should support train, val, and test splits.](https://github.com/vanvalenlab/deepcell-tf/pull/512#discussion_r625955276)
- [The bounds on `deepcell.data.tracking.random_translate` should be confirmed to be valid.](https://github.com/vanvalenlab/deepcell-tf/pull/512#discussion_r626243892)

willgraf:
I think this PR is very close to being ready, but we need to update the example notebooks.

willgraf:
In addition to tests and an updated notebook, the trained model in `deepcell.applications` has to be updated.

There has been model changes since the current version was published and it can no longer be loaded properly.

willgraf:
#532 addresses much of the required testing and fixed a bug in the `temporal_slice` function (Thanks @R-Ding!)

The remaining blockers for this PR (in my view) are:

- [ ] Cleaned up notebook for training/running the model.
- [ ] Update model architecture (@vanvalen found the temporal convolution layers cause data leaking)
- [ ] Updated tracking application with trained model (using new architecture)
- [ ] [The deepcell.data.split_data function should support train, val, and test splits.](https://github.com/vanvalenlab/deepcell-tf/pull/512#discussion_r625955276)
- [ ] [The bounds on deepcell.data.tracking.random_translate should be confirmed to be valid.](https://github.com/vanvalenlab/deepcell-tf/pull/512#discussion_r626243892)

willgraf (commented):
This is looking really good IMO. I think the code is pretty much ready (although it would be great to make the model easier to understand). I'd like to run through the notebook again to:

- [x] get good benchmarking results at the end of it
- [x] remove/fix warnings for the gif file section
- [x] clarify the `1` index that we are doing in the "Benchmark" section
- [x] use augmentation

---

I also want to verify the image mpp for the training set, as we have it set to 0.65, but I think 0.5 is normal for 20x?

MekWarrior:
I believe MPP for tracking model used in the application is 0.65.

willgraf (approved):

willgraf (approved):

PR #510, opened 2021-04-12
Graph work

ulisrael:
## What
* created test_graph_utils.py 

## Why
* setting up test infra for graph functions


deepcell/utils/data_utils.py
```diff
@@ -507,3 +509,28 @@ def reshape_movie(X, y, reshape_size=256):
...
+    
+    return data + np.array([[t,t]])
+
+def reflection_mat2D(data):
+    """apply 2D reflection"""
+    assert data.shape[-1] ==2, 'data is not 2D'
+    theta = np.radians(np.random.randint(0,360))
+    ref = np.array([[np.cos(2*theta), np.sin(2*theta)],
+                    [np.sin(2*theta), -np.cos(2*theta)]])
+    return tf.matmul(data,ref)
```
willgraf, line 536:
```suggestion
    return tf.matmul(data,ref)

```

Dockerfile
```diff
@@ -6,9 +6,17 @@ ARG TF_VERSION=2.4.1-gpu
...
+    python3-tk \
+    graphviz \
+    libxext6 \
+    libxrender-dev \
+    libsm6 && \
+    rm -rf /var/lib/apt/lists/* && \
+    /usr/bin/python3 -m pip install --upgrade pip
 
-WORKDIR /notebooks
+#WORKDIR notebooks/
```
willgraf, line 19: Let's revert this.

```suggestion
WORKDIR /notebooks
```

Dockerfile
```diff
@@ -23,6 +31,9 @@ COPY deepcell /opt/deepcell-tf/deepcell
 # Install deepcell via setup.py
 RUN pip install /opt/deepcell-tf
 
+## need to remove jedi - causes erros for tab in jupyter
+RUN pip uninstall jedi --yes
+
```
willgraf, line 36: What installs `jedi`? If possible it would be good to prevent this from getting installed since more Docker layers means larger Docker images.

deepcell/utils/data_utils.py
```diff
@@ -507,3 +509,28 @@ def reshape_movie(X, y, reshape_size=256):
...
+
+
+def rotation_mat2D(data):
+    """apply 2D roation"""
+    assert data.shape[-1] == 2, 'data is not 2D'
+    theta = np.radians(np.random.randint(0,360))
+    rot = np.array([[np.cos(theta), -np.sin(theta)],
+                    [np.sin(theta), np.cos(theta)]])
+    
+    return tf.matmul(data,rot)
```
willgraf, line 521: Could this be `np.matmul` instead? Mixing the np.arrays with tensorflow makes me a bit nervous.

requirements.txt
```diff
@@ -1,8 +1,15 @@
...
 jupyter>=1.0.0,<2
 deepcell-tracking>=0.3.1,<0.4.0
 deepcell-toolbox>=0.9.0,<0.10.0
+opencv-python<=3.4.9.31
+spektral==1.0.6
+pydot>=1.4.1
+pydotplus>=2.0.2
+graphviz>=0.14
+einops==0.3.0
+jupyterlab
```
willgraf, line 15: I'd prefer to let users install `jupyterlab` if they want it, but not to ship with it by default.

```suggestion
```

requirements.txt
```diff
@@ -1,8 +1,15 @@
 numpy>=1.16.6,<1.20
 scipy>=1.2.3,<2
-scikit-image>=0.14.5
+scikit-image==0.16.1
 scikit-learn>=0.20.4
 tensorflow==2.4.1
 jupyter>=1.0.0,<2
 deepcell-tracking>=0.3.1,<0.4.0
 deepcell-toolbox>=0.9.0,<0.10.0
+opencv-python<=3.4.9.31
```
willgraf, line 9: Let's use `opencv-python-headless` instead. I believe it's also installed via the `deepcell-toolbox`, so please make sure the version restrictions are compatible.

deepcell/utils/data_utils.py
willgraf: Here too, could we use `np.matmul` instead?

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+
+def get_max_cells(label_image):
+    """ Compute the maximum number of cells in a single batch for a label image
+
+    Args:
+        label_image (int): Label image where every cell has been given a
+            unique integer id
+
+    Returns:
+        np.int: The maximum number of cells in a single batch of the
```
willgraf, line 48: I think its just a regular `int`.
```suggestion
        int: The maximum number of cells in a single batch of the
```

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+                   distance_threshold=0.5,
+                   self_connection=True):
+    """Convert a label image to a graph
+
+    Args:
+        label_image (int): Label image where every cell has been given a
+            unique integer id
+        distance_threshold (float): Connect two cells with an edge if their
+            centroids are closer than distance_threshold pixels
+        self_connection (bool): keep or remove diagonal of adj matrix
```
willgraf, line 70: Some minor docs cleanup

```suggestion
        label_image (int): Label image where every object has been given a
            unique integer id
        distance_threshold (float): Connect two cells with an edge if their
            centroids are closer than ``distance_threshold`` pixels
        self_connection (bool): keep or remove diagonal of adj matrix
```

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Instantiate the adjacency matrix
+    adjacency_matrix = np.zeros((label_image.shape[0], max_cells, max_cells))
+    centroid_matrix = np.zeros((label_image.shape[0], max_cells, 2))
+    label_matrix = np.zeros((label_image.shape[0], max_cells))
+
+    for batch in range(label_image.shape[0]):
+        label_image_batch = label_image[batch]
+
+        props = regionprops_table(label_image_batch, properties=['centroid', 'label'])
```
willgraf, line 92: why `regionprops_table` over `regionprops`?

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+    if image.shape[0] != label_image.shape[0]:
+        raise ValueError("image and label_image must have the same size for "
+                         "the batch dimmension")
+
+    # Find out the max number of cells
+    max_cells = get_max_cells(label_image)
+
+    # Instantiate the feature matrix
+    feature_matrix = np.zeros((image.shape[0], max_cells, image.shape[-1]))
+    label_matrix = np.zeros((label_image.shape[0], max_cells))
```
willgraf, line 137: It looks like the `label_matrix` is not used for anything.

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+            mean_intensity = props['mean_intensity']
+            labels = props['label']
+
+            feature_matrix[batch, 0:mean_intensity.shape[0], feature] = mean_intensity
+            label_matrix[batch, 0:labels.shape[0]] = labels
+
+    return feature_matrix
+
+
+def get_celltypes(label_image, celltype_image):
```
willgraf, line 158: Let's refactor `celltype_image` to `classification_image` or `semantic_class_image` or something that is not specific to cell types.

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+
+    for batch in range(label_image.shape[0]):
+        label_image_batch = np.squeeze(label_image[batch])
+        celltype_image_batch = np.squeeze(celltype_image[batch])
+
+        props = regionprops_table(label_image_batch,
+                                  properties=['coords', 'label'])
+
+        for i in range(len(props['coords'])):
+            coords = props['coords'][i]
```
willgraf, line 193: You can iterate over both at once using `enumerate`

```suggestion
        for i, coords in enumerate(props['coords']):
```

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+
+    return celltype_matrix
+
+
+def adj_to_degree(adj, power=-0.5, epsilon=1e-5):
+    """ Convert adjacency matrix to degree matrix
+
+    Args:
+        adj (float): Adjacency matrix
+        power (float): tbd
```
willgraf, line 206: We should describe what this is.

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+    d_mat_inv_sqrt = adj_to_degree(adj, power=-0.5)
+
+    # Compute D^-0.5AD^0.5 - Recall that D and A are symmetric
+    normalized_adjacency = np.matmul(d_mat_inv_sqrt, adj)
+    normalized_adjacency = np.matmul(normalized_adjacency, d_mat_inv_sqrt)
+
+    return normalized_adjacency
+
+
+def preprocess_adj(adj):
```
willgraf, line 246: Can we rename this function to be more descriptive? Maybe something like `convert_matrix_to_tuple`? Ideally it will be easy to know what's happening to the data from the function name.

deepcell/graph_utils.py
```diff
@@ -0,0 +1,274 @@
...
+        np.array: Symmetrically normalized adjacency matrix after adding self
+            connections to each node
+    """
+    adj_normalized = normalize_adj(adj + sp.eye(adj.shape[0]))
+
+    return adj_normalized
+
+
+def node_dist(prop0, prop1):
+    """ Compute minimum distance between two cells
```
willgraf, line 263:
```suggestion
    """ Compute minimum distance between two objects
```

deepcell/graph_utils_tests.py
```diff
@@ -0,0 +1,290 @@
...
+
+    Args:
+        adj: Adjacency matrix
+
+    Returns:
+        np.array: Degree matrix raised to a given power
+    """
+
+
+def test_normalize_adj(adj):
```
willgraf, line 257: This should probably have a test.

deepcell/graph_utils_tests.py
```diff
@@ -0,0 +1,290 @@
...
+    y = np.expand_dims(y,axis=-1)
+
+    # true cell types
+    true_celltypes = np.array([2., 2., 0., 0., 1.])
+
+    celltype_matrix = graph_utils.get_celltypes(x, y)
+
+    assert (true_celltypes == celltype_matrix).all()
+
+def test_adj_to_degree(adj, power=-0.5, epsilon=1e-5):
```
willgraf, line 246: This should probably have a test.

willgraf (changes requested):
A good start! Please see the comments. Also, let's remove some of the more specific code from this PR:
- [ ] jupyter notebooks that do not serve as examples of how to use the code.
- [ ] helper functions that are specific for a single datasets.

willgraf:
I'm going to close this PR since there has been no activity for 9 months. Please re-open when it is ready/being worked on.

PR #532, opened 2021-07-02
Tests and bug fix for deepcell.data.tracking

R-Ding:
## What
* Add tests for `deepcell.data.tracking`.
* Fix bug in `deepcell.data.tracking.temporal_slice` function.

## Why
* Tests help to ensure data is being prepared correctly.
* Temporal dimension when `temporal_slice` is called is at index 0 instead of 1 because the data has not been batched yet.


deepcell/data/tracking_test.py
```diff
@@ -0,0 +1,180 @@
...
+        self.assertEqual(rotated_X['centroids'].shape, X_cents.shape)
+
+        # Assert rotating by 0 does not change tensors
+        X_apps = rotated_X['appearances'][0, 0, 0, 0]
+        X_cents = rotated_X['centroids'][0, 0, 0]
+        output_X, y = tracking.random_rotate(rotated_X, y, 0)
+        r0 = (float)(output_X['appearances'][0, 0, 0, 0] - X_apps)
+        self.assertEqual(r0, 0)
+        r0 = (float)(output_X['centroids'][0, 0, 0] - X_cents)
+        self.assertEqual(r0, 0)
```
willgraf, line 105: It looks like you are testing that a single value of the X arrays are unchanged.

Instead, check that the entire array is the same with the [`assertAllEqual` testing helper](https://www.tensorflow.org/api_docs/python/tf/test/TestCase).

Something like:

```python
X_apps = rotated_X['appearances']
X_cents = rotated_X['centroids']

output_X, y = tracking.random_rotate(rotated_X, y, 0)

self.assertAllEqual(output_X['appearances'], X_apps)
self.assertAllEqual(output_X['centroids'], X_cents)
```

deepcell/data/tracking_test.py
```diff
@@ -0,0 +1,180 @@
...
+        r0 = (float)(translated_X['centroids'][0, 0, 0] - init_x_0)
+        r1 = (float)(translated_X['centroids'][0, 1, 0] - init_x_1)
+        # Assert that centroids are translated by the same amount
+        assert abs(r1 - r0) < 0.001
+
+        # Assert range of 0 does not translate data
+        X_cents = translated_X['centroids'][0, 0, 0]
+        output_X, y = tracking.random_translate(X, y, range=0)
+        r0 = (float)(output_X['centroids'][0, 0, 0] - X_cents)
+        self.assertEqual(r0, 0)
```
willgraf, line 127: Again, we can probably test the entire array is unchanged here.

willgraf (changes requested):

deepcell/data/tracking_test.py
```diff
@@ -0,0 +1,179 @@
...
+        init_x_0 = X['centroids'][0, 0, 0]
+        init_x_1 = X['centroids'][0, 1, 0]
+
+        translated_X, y = tracking.random_translate(X, y)
+
+        # Get distance translated for two different cells
+        r0 = (float)(translated_X['centroids'][0, 0, 0] - init_x_0)
+        r1 = (float)(translated_X['centroids'][0, 1, 0] - init_x_1)
+        # Assert that centroids are translated by the same amount
+        assert abs(r1 - r0) < 0.001
```
willgraf, line 121: Not a bad start, but a few things I think we can improve:

* The tolerance of 1e-3 seems rather arbitrary. Shouldn't they get translated the exact same amount?
* We can make this assertion for all points in the array, not just 2 random ones.

Here's a quick mockup of how I think these can be addressed:

```python
translated_X, y = tracking.random_translate(X, y)

diff = X['centroids'] - translated_X['centroids']

# each value should be translated the same amount
single_value = np.empty_like(diff)
single_value.fill(diff[0, 0, 0])
self.assertAllEqual(diff, single_value)
```

willgraf (changes requested):

willgraf (approved):
Some good tests for the tracking datasets! LGTM.

PR #535, opened 2021-07-20
Cache the entire Python environment to speed up build times.

willgraf:
## What
* Cache the entire Python environment in the testing GitHub Action workflow.

## Why
* [Drastically speed up build times](https://medium.com/ai2-blog/python-caching-in-github-actions-e9452698e98d).


willgraf:
The first build (no cache) with these changes took 9m 24s.

All jobs were restarted (to use the new cache) and which took 6m 14s.

---

Installing dependencies went from ~1:43 to ~0:20.

PR #536, opened 2021-07-23
Graph tracking model architecture changes

R-Ding:
## What
* Create keras layer `Unmerge` used to unmerge embeddings and centroids.
* Change default `merge_type` for `get_embedding_temporal_merge_model` and `get_delta_temporal_merge_model` to lstm instead of cnn
* Remove residual connections in `get_tracking_decoder` 

## Why
* Nicer to have a keras layer than use Lambda layers
* Merging the embeddings depends only on timepoints in the present and past, and thus do not have information from future embeddings. Cnn's use information from future timepoints, resulting in temporal information being leaked. Lstm's only use information from the current and old embeddings (which is what we want).
* Adding the residual connections does not help with accuracy

### DIFF
--- a/deepcell/model_zoo/tracking.py
+++ b/deepcell/model_zoo/tracking.py
@@ -233,12 +233,35 @@ class DeltaReshape(tf.keras.layers.Layer):
         return output
 
 
+class Unmerge(tf.keras.layers.Layer):
+    def __init__(self, track_length, max_cells, embedding_dim, **kwargs):
+        super(Unmerge, self).__init__(**kwargs)
+        self.track_length = track_length
+        self.max_cells = max_cells
+        self.embedding_dim = embedding_dim
+
+    def call(self, inputs):
+        new_shape = [-1, self.track_length, self.max_cells, self.embedding_dim]
+        output = tf.reshape(inputs, new_shape)
+
+        return output
+
+    def get_config(self):
+        config = {
+            'track_length': self.track_length,
+            'max_cells': self.max_cells,
+            'embedding_dim': self.embedding_dim
+        }
+        base_config = super(Unmerge, self).get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
+
 class GNNTrackingModel(object):
     """Creates a tracking model based on Graph Neural Networks(GNNs).
 
     Args:
         n_filters (int): Number of filters
-        endcoder_dim (int): Dimension of encoder
+        encoder_dim (int): Dimension of encoder
         embedding_dim (int): Dimension of embedding
         n_layers (int): number of layers
         time_window (int): number of frames to include in temporal merges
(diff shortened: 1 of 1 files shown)


deepcell/model_zoo/tracking.py
```diff
@@ -351,7 +364,7 @@ def get_appearance_encoder(self):
         x = TimeDistributed(ImageNormalization2D(norm_method='whole_image',
                                                  name='imgnrm_ae'))(x)
 
-        for i in range(self.time_window):
+        for i in range(5):
```
willgraf, line 367: Any reason to hardcode 5 rather than leave it as a configurable parameter?

deepcell/model_zoo/tracking.py
```diff
@@ -287,7 +310,7 @@ def __init__(self,
         # Create model
         self.training_model, self.inference_model = self.get_models()
 
-    def get_embedding_temporal_merge_model(self, merge_type='cnn'):
+    def get_embedding_temporal_merge_model(self, merge_type='lstm'):
```
willgraf, line 313: If only LSTM works, let's just remove the "cnn" option and `merge_type` parameter entirely.

willgraf: Same with `get_delta_temporal_merge_model`.

willgraf (changes requested):

deepcell/model_zoo/tracking.py
R-Ding: Dave’s reasoning is we only want to do 5 convolution and pooling operations and it’s alright to have it hardcoded for now as the crop size (32) is also currently hardcoded.

willgraf (approved):
I'm a little uncomfortable with hardcoding the the conv/pool ops as well as hardcoding the appearance shape. But the appearance shape is hardcoded in the main branch so we can resolve that before merging the PR.

I'd like @MekWarrior to review and approve as well, but this looks like a good improvement.

deepcell/model_zoo/tracking.py
```diff
@@ -233,12 +233,35 @@ def call(self, inputs):
         return output
 
 
+class Unmerge(tf.keras.layers.Layer):
```
MekWarrior, line 236: Should this be moved to the `layers` module?

MekWarrior (approved):
I agree with Will that it strikes me as odd to hardcode something when the variable is already there. But If it works, it's better than our current version.

deepcell/model_zoo/tracking.py
R-Ding: I was wondering that too, there's also two other layers defined here, `Comparison` and `DeltaReshape`. Should we move all three into an existing `layers` file or a new one?

willgraf: We should probably move them into a new layers file. IDK what to name it, probably should be more about the temporal/graph nature rather than tracking.

Each of them should have tests as well, but they cannot use the standard testing routine since they have multiple inputs.

I think the `FilterDetections` layer has some good tests that are not using the `layer_test` method.

MekWarrior: We have  `temporal`set up already - that may be a good fit.

willgraf: I'm going to merge this PR and update the TODO checklist to include migrating and testing the layers.

PR #537, opened 2021-07-26
Add support for test split in `deepcell.data.split_data`

R-Ding:
## What
* Allow `split_dataset` to accept test data fraction. 
* In `prepare_dataset, add `test_data` when unpacking returns from `split_dataset` call, and properly batch/prefetch the test data.
* In `tracking_test`, modify to include `test_split` parameters, remove third test case, use `tf.random.uniform` to create test data.

## Why
* Want to allow users to split data into test data in addition to training and validation data.
* Third test case makes tests too long
* `tf.random.uniform` more concise and clear than previous implementation

### DIFF
--- a/deepcell/data/__init__.py
+++ b/deepcell/data/__init__.py
@@ -30,33 +30,49 @@ from __future__ import division
 from __future__ import print_function
 
 
-def split_dataset(dataset, validation_data_fraction):
+def split_dataset(dataset, val_data_fraction, test_data_fraction=0):
     """
-    Splits a dataset of type tf.data.Dataset into a training and validation
-    dataset using given ratio. Fractions are rounded up to two decimal places.
+    Splits a dataset of type tf.data.Dataset into a training, validation, and
+    optionally test dataset using given ratios. Fractions are rounded up to
+    two decimal places.
 
     Inspired by: https://stackoverflow.com/a/59696126
 
     Args:
         dataset (tf.data.Dataset): the input dataset to split.
-        validation_data_fraction (float): the fraction of the validation data
+        val_data_fraction (float): the fraction of the validation data
             between 0 and 1.
+        test_data_fraction (float): the fraction of the test data between 0
+            and 1.
 
     Returns:
-        (tf.data.Dataset, tf.data.Dataset): a tuple of (training, validation).
+        (tf.data.Dataset, tf.data.Dataset, tf.data.Dataset): a tuple of
+            (training, validation, test).
     """
-    validation_data_percent = round(validation_data_fraction * 100)
-    if not (0 <= validation_data_percent <= 100):
-        raise ValueError('validation_data_fraction must be ∈ [0,1].')
+    val_data_percent = round(val_data_fraction * 100)
+    if not 0 <= val_data_percent <= 100:
+        raise ValueError('val_data_fraction must be ∈ [0,1].')
+
+    test_data_percent = round(test_data_fraction * 100)
+    if not 0 <= test_data_percent <= 100:
+        raise ValueError('test_data_fraction must be ∈ [0,1].')
+
+    if val_data_percent + test_data_percent >= 100:
+        raise ValueError('sum of val_data_fraction and '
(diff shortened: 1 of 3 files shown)


deepcell/data/__init__.py
```diff
@@ -30,33 +30,49 @@
...
+    if not 0 <= val_data_percent <= 100:
+        raise ValueError('val_data_fraction must be ∈ [0,1].')
+
+    test_data_percent = round(test_data_fraction * 100)
+    if not 0 <= test_data_percent <= 100:
+        raise ValueError('test_data_fraction must be ∈ [0,1].')
+
+    if val_data_percent + test_data_percent >= 100:
+        raise ValueError('sum of val_data_fraction and '
+                         + 'test_data_fraction must be ∈ [0,1].')
```
willgraf, line 62: There is no need to concatenate these strings, as they are inside the parentheses. This will be inferred as a multi-line string:

```suggestion
        raise ValueError('sum of val_data_fraction and '
                         'test_data_fraction must be ∈ [0,1].')
```

deepcell/data/tracking_test.py
```diff
@@ -129,52 +125,51 @@ def test_random_translate(self, time, max_cells, crop_size,
...
+                                                                   seed=seed,
+                                                                   track_length=track_length,
+                                                                   val_split=val_split,
+                                                                   test_split=test_split)
         # Test data correctly batched
         for X, y in train_data.take(1):
             self.assertEqual((X['appearances'].shape)[0], batch_size)
         for X, y in val_data.take(1):
             self.assertEqual((X['appearances'].shape)[0], batch_size)
+        if not test_split == 0:
```
willgraf, line 144: Slightly clearer

```suggestion
        if test_split != 0:
```

willgraf: What happens if it is 0? Will there be an exception raised? Let's add a test for the `else` clause here.

willgraf (changes requested):
Looks pretty good, just some small touchups.

deepcell/data/tracking_test.py
MekWarrior: @R-Ding This is especially important as we specifically set `test_split = 0` as default

R-Ding: I am having trouble finding a way to test whether `test_data` is empty when `test_split=0`, especially because after applying `.filter()` to any dataset, the cardinality becomes -2. Do either of you have any suggestions?

deepcell/data/__init__.py
```diff
@@ -30,33 +30,49 @@
...
+                         'test_data_fraction must be ∈ [0,1].')
 
     dataset = dataset.enumerate()
-    train_dataset = dataset.filter(lambda f, data: f % 100 > validation_data_percent)
-    validation_dataset = dataset.filter(lambda f, data: f % 100 <= validation_data_percent)
+    val_dataset = dataset.filter(lambda f, data: f % 100 <= val_data_percent)
+    train_dataset = dataset.filter(lambda f, data:
+                                   f % 100 > test_data_percent + val_data_percent)
+    test_dataset = dataset.filter(lambda f, data: f % 100 > val_data_percent and
+                                  f % 100 <= val_data_percent + test_data_percent)
```
willgraf, line 69: Can you add a comment/TODO that this will cause problems if there are fewer than 100 records?

willgraf (changes requested):
Just some last-minute name change suggestions and some comments.

deepcell/data/__init__.py
```diff
@@ -30,33 +30,50 @@
...
 
     Inspired by: https://stackoverflow.com/a/59696126
 
     Args:
         dataset (tf.data.Dataset): the input dataset to split.
-        validation_data_fraction (float): the fraction of the validation data
+        val__size (float): the fraction of the validation data
             between 0 and 1.
+        test__size (float): the fraction of the test data between 0
+            and 1.
```
willgraf, line 46: Looks like there's an extra underscore in there

willgraf (approved):
LGTM!

PR #540, opened 2021-08-07
Application batch prediction

### LINKED ISSUE #538: Memory error using large images with application
**Describe the bug**
The Mesmer application is OOMing when predicting on large images. Given that we tile the inputs to the model and batch process the data, my suspicion is that it's the pre or post processing that is causing the problem. Will need to investigate to determine which step in the application code is causing the error. 

**To Reproduce**
Run app.predict on 18000x21000 image

ngreenwald:
## What
Modifies the applications to use internal batch prediction function rather than the default model.predict batching functionality. Closes #538 

## Why
The default model.predict batch function creates a tf.dataset object with all images, not just the specified batch size. This leads to memory issues on the GPU when batch processing tiles from large images.

### DIFF
--- a/deepcell/applications/application.py
+++ b/deepcell/applications/application.py
@@ -312,9 +312,7 @@ class Application(object):
 
         return image
 
-    def _batch_predict(self,
-                       tiles,
-                       batch_size):
+    def _batch_predict(self, tiles, batch_size):
         """Batch process tiles to generate model predictions.
 
         The built-in keras.predict function has support for batching, but loads the entire
@@ -324,7 +322,6 @@ class Application(object):
         Args:
             tiles: Tiled data which will be fed to model
             batch_size: Number of images to predict on per batch
-
 
         Returns:
             list: Model outputs
@@ -339,17 +336,17 @@ class Application(object):
         output_tiles = []
 
         # loop through each batch
-        for batch_index, (batch_start, batch_end) in enumerate(batches):
-            batch_inputs = tiles[batch_start:batch_end, ...]
+        for i in range(0, tiles.shape[0], batch_size):
+            batch_inputs = tiles[i:i + batch_size, ...]
 
             batch_outputs = self.model.predict(batch_inputs, batch_size=batch_size)
 
             # model with only a single output gets temporarily converted to a list
-            if type(batch_outputs) != list:
+            if not isinstance(batch_outputs, list):
                 batch_outputs = [batch_outputs]
 
             # initialize output list with empty arrays to hold all batches
-            if batch_index == 0:
+            if not output_tiles:
                 for batch_out in batch_outputs:
                     shape = (num_images,) + batch_out.shape[1:]
                     output_tiles.append(np.zeros(shape, dtype=tiles.dtype))
--- a/deepcell/applications/application.py
+++ b/deepcell/applications/application.py
@@ -321,7 +321,7 @@ class Application(object):
         model.predict function without soaking up GPU memory.
 
         Args:
-            tiles (np.array): Tiled data which will be fed to model
+            tiles (numpy.array): Tiled data which will be fed to model
             batch_size (int): Number of images to predict on per batch
 
         Returns:
--- a/deepcell/applications/application.py
+++ b/deepcell/applications/application.py
@@ -321,8 +321,8 @@ class Application(object):
         model.predict function without soaking up GPU memory.
 
         Args:
-            tiles: Tiled data which will be fed to model
-            batch_size: Number of images to predict on per batch
+            tiles (np.array): Tiled data which will be fed to model
+            batch_size (int): Number of images to predict on per batch
 
         Returns:
             list: Model outputs
--- a/deepcell/applications/application.py
+++ b/deepcell/applications/application.py
@@ -327,7 +327,7 @@ class Application(object):
 
 
         Returns:
-            np.array or list: Model outputs
+            list: Model outputs
         """
 
         num_images = tiles.shape[0]
--- a/deepcell/applications/application.py
+++ b/deepcell/applications/application.py
@@ -312,6 +312,54 @@ class Application(object):
 
         return image
 
+    def _batch_predict(self,
+                       tiles,
+                       batch_size):
+        """Batch process tiles to generate model predictions.
+
+        The built-in keras.predict function has support for batching, but loads the entire
+        image stack into GPU memory, which is prohibitive for large images. This function uses
+        similar code to the underlying model.predict function without soaking up GPU memory.
+
+        Args:
+            tiles: Tiled data which will be fed to model
+            batch_size: Number of images to predict on per batch
+
+
+        Returns:
+            np.array or list: Model outputs
+        """
+
+        num_images = tiles.shape[0]
+        num_batches = int(np.ceil(num_images / float(batch_size)))
+        batches = [(i * batch_size, min(num_images, (i + 1) * batch_size)) for i in
+                   range(0, num_batches)]
+
+        # list to hold final output
+        output_tiles = []
+
+        # loop through each batch
+        for batch_index, (batch_start, batch_end) in enumerate(batches):
+            batch_inputs = tiles[batch_start:batch_end, ...]
+
+            batch_outputs = self.model.predict(batch_inputs, batch_size=batch_size)
+
+            # model with only a single output gets temporarily converted to a list
+            if type(batch_outputs) != list:
+                batch_outputs = [batch_outputs]
+
+            # initialize output list with empty arrays to hold all batches
+            if batch_index == 0:
+                for batch_out in batch_outputs:
+                    shape = (num_images,) + batch_out.shape[1:]
+                    output_tiles.append(np.zeros(shape, dtype=tiles.dtype))
+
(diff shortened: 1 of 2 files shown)


deepcell/applications/application.py
```diff
@@ -312,6 +312,54 @@ def _resize_output(self, image, original_shape):
...
+        num_batches = int(np.ceil(num_images / float(batch_size)))
+        batches = [(i * batch_size, min(num_images, (i + 1) * batch_size)) for i in
+                   range(0, num_batches)]
+
+        # list to hold final output
+        output_tiles = []
+
+        # loop through each batch
+        for batch_index, (batch_start, batch_end) in enumerate(batches):
+            batch_inputs = tiles[batch_start:batch_end, ...]
```
willgraf, line 343: I think this would be much clearer if you just use the step size with the `range` function:

```suggestion
        for i in range(0, tiles.shape[0], batch_size):
            batch_inputs = tiles[i:i + batch_size, ...]
```

deepcell/applications/application.py
```diff
@@ -312,6 +312,54 @@ def _resize_output(self, image, original_shape):
...
+
+            # model with only a single output gets temporarily converted to a list
+            if type(batch_outputs) != list:
+                batch_outputs = [batch_outputs]
+
+            # initialize output list with empty arrays to hold all batches
+            if batch_index == 0:
+                for batch_out in batch_outputs:
+                    shape = (num_images,) + batch_out.shape[1:]
+                    output_tiles.append(np.zeros(shape, dtype=tiles.dtype))
```
willgraf, line 355: Since `output_tiles` is a list, we can just do a truthy-ness check to verify if it needs to be initialized:

```suggestion
            # initialize output list with empty arrays to hold all batches
            if not output_tiles:
                for batch_out in batch_outputs:
                    shape = (num_images,) + batch_out.shape[1:]
                    output_tiles.append(np.zeros(shape, dtype=tiles.dtype))
```

deepcell/applications/application.py
```diff
@@ -312,6 +312,54 @@ def _resize_output(self, image, original_shape):
...
+        output_tiles = []
+
+        # loop through each batch
+        for batch_index, (batch_start, batch_end) in enumerate(batches):
+            batch_inputs = tiles[batch_start:batch_end, ...]
+
+            batch_outputs = self.model.predict(batch_inputs, batch_size=batch_size)
+
+            # model with only a single output gets temporarily converted to a list
+            if type(batch_outputs) != list:
```
willgraf, line 348: Prefer `isinstance` over `type` comparisons:

```suggestion
            if not isinstance(batch_outputs, list):
```

deepcell/applications/application.py
```diff
@@ -312,6 +312,54 @@ def _resize_output(self, image, original_shape):
...
         return image
 
+    def _batch_predict(self,
+                       tiles,
+                       batch_size):
+        """Batch process tiles to generate model predictions.
+
+        The built-in keras.predict function has support for batching, but loads the entire
+        image stack into GPU memory, which is prohibitive for large images. This function uses
+        similar code to the underlying model.predict function without soaking up GPU memory.
```
willgraf, line 322: Can you please shorten the line lengths to ideally <80?

deepcell/applications/application_test.py
```diff
@@ -243,13 +243,27 @@ def _format_model_output(Lx):
...
 
+    def test_batch_predict(self):
+        for num_images in [4, 8, 10]:
+            for num_pred_heads in [1, 2]:
+                model = DummyModel(n_out=num_pred_heads)
+                app = Application(model)
+
+                x = np.random.rand(num_images, 128, 128, 1)
+                y = app._batch_predict(x, batch_size=4)
+                self.assertEqual(x.shape, y[0].shape)
```
willgraf, line 254: I think a good test here would be to use [`pytest-mock` to "spy"](https://github.com/pytest-dev/pytest-mock/#spy) on `model.predict` (or another function) and make sure it's called the correct number of times.

deepcell/applications/application.py
```diff
@@ -312,6 +312,54 @@ def _resize_output(self, image, original_shape):
 
         return image
 
+    def _batch_predict(self,
+                       tiles,
+                       batch_size):
```
willgraf, line 317: No need for the newlines

```suggestion
    def _batch_predict(self, tiles, batch_size):
```

deepcell/applications/application.py
```diff
@@ -312,6 +312,54 @@ def _resize_output(self, image, original_shape):
...
+        """Batch process tiles to generate model predictions.
+
+        The built-in keras.predict function has support for batching, but loads the entire
+        image stack into GPU memory, which is prohibitive for large images. This function uses
+        similar code to the underlying model.predict function without soaking up GPU memory.
+
+        Args:
+            tiles: Tiled data which will be fed to model
+            batch_size: Number of images to predict on per batch
+
```
willgraf, line 327: Remove extra whitespace

```suggestion
```

willgraf (changes requested):

deepcell/applications/application_test.py
ngreenwald: I'm running into issues getting mocker declared in the test function, since it's a method of a test class. 
When I do 
```
    def test_batch_predict(self, mocker):
        for num_images in [4, 8, 10]: 
```

I get the following error message about a missing argument in test_batch_predict when pytest is trying to call the method. I'm guessing I need to use a decorator or something instead?

image

willgraf: Oh, I forgot that TensorFlow uses `unittest` instead. Try using [the `mock` library](https://docs.python.org/3/library/unittest.mock.html#the-mock-class) instead

deepcell/applications/application.py
```diff
@@ -312,6 +312,47 @@ def _resize_output(self, image, original_shape):
...
+        """Batch process tiles to generate model predictions.
+
+        The built-in keras.predict function has support for batching, but
+        loads the entire image stack into GPU memory, which is prohibitive
+        for large images. This function uses similar code to the underlying
+        model.predict function without soaking up GPU memory.
+
+        Args:
+            tiles: Tiled data which will be fed to model
+            batch_size: Number of images to predict on per batch
```
willgraf, line 325: Don't forget the data type for the args!

```suggestion
            tiles (np.array): Tiled data which will be fed to model
            batch_size (int): Number of images to predict on per batch
```

deepcell/applications/application_test.py
```diff
@@ -243,13 +244,43 @@ def _format_model_output(Lx):
...
+                x = np.random.rand(num_images, 128, 128, 1)
+
+                if num_pred_heads == 1:
+                    app.model.predict = Mock(side_effect=predict1)
+                else:
+                    app.model.predict = Mock(side_effect=predict2)
+                y = app._batch_predict(x, batch_size=4)
+                self.assertEqual(x.shape, y[0].shape)
+
+                assert app.model.predict.call_count == np.ceil(num_images / 4)
```
willgraf, line 271: Please set batch_size equal to a variable and include in the `np.ceil(num_images / batch_size)`?

willgraf (changes requested):
Looks really good, just some last finishing touches.

deepcell/applications/application.py
```diff
@@ -321,8 +321,8 @@ def _batch_predict(self, tiles, batch_size):
         model.predict function without soaking up GPU memory.
 
         Args:
-            tiles: Tiled data which will be fed to model
-            batch_size: Number of images to predict on per batch
+            tiles (np.array): Tiled data which will be fed to model
```
willgraf, line 324: Looks like I was wrong and it should be `numpy.array` instead.

deepcell/applications/application_test.py
```diff
@@ -34,6 +34,7 @@
 from tensorflow.keras.layers import Input
 from tensorflow.python.platform import test
 from unittest.mock import Mock
+from itertools import product
```
willgraf, line 37: This should actually go in its own block, between the future imports and the 3rd party imports. (same with `unittest`):

```python
from __future__ import absolute_import
from __future__ import division

from itertools import product
from unittest.mock import Mock

import numpy as np
from tensorflow.keras.layers import Input
from tensorflow.python.platform import test
```

deepcell/applications/application_test.py
```diff
@@ -29,9 +29,10 @@
 from __future__ import division
 from __future__ import print_function
 
+from itertools import product
 import numpy as np
+from unittest.mock import Mock
```
willgraf, line 34:
```suggestion
from itertools import product
from unittest.mock import Mock

import numpy as np
```

willgraf (changes requested):
One more nitpicky thing

deepcell/applications/application_test.py
ngreenwald: Just so that imports with the same structure are next to each other? Or because we're importing numpy with an abbreviation?

willgraf: It's a convention to group imports by future, built-in, 3rd party, custom/local module imports.

willgraf (approved):
LGTM

PR #545, opened 2021-08-14
Support individual parameter passing to Mesmer app

ngreenwald:
## What
Previously, the default values for the Mesmer post-processing kwargs were configured such that if the user passed any args to the app, all of the other defaults would be reset. This PR modifies the behavior so that only the kwargs specified by the user are changed, the other defaults remain unchanged

## Why
Removes confusing behavior where users would think they're only overriding a single arg, when in fact they would be resetting all of the args back to the `deep_watershed` defaults, which are not the same as the `Mesmer` defaults


willgraf (approved):
LGTM

deepcell/applications/mesmer_test.py
```diff
@@ -222,6 +224,47 @@ def test_mesmer_app(self):
...
+                'interior_smooth': 2,
+                'small_objects_threshold': 15,
+                'fill_holes_threshold': 15,
+                'radius': 2
+            }
+
+            # all defaults
+            _ = app.predict(x, compartment='whole-cell')
+            args = app._predict_segmentation.call_args[1]
+            assert args['postprocess_kwargs']['whole_cell_kwargs'] == default_kwargs
```
willgraf, line 243: I don't like that if we change the defaults in the application, we have to change the test. Let's just get the defaults from it and not make this assertion.

ngreenwald: Should I put them somewhere else so they can be imported?

willgraf: I think you can access them just as you did before:

```python
args = app._predict_segmentation.call_args[1]
default_kwargs = args['postprocess_kwargs']['whole_cell_kwargs']
```

Then you can make sure that the rest of wholecell/nuclear kwargs are equal to the defaults, except the new keys that were overridden.

deepcell/applications/mesmer_test.py
```diff
@@ -222,6 +224,38 @@ def test_mesmer_app(self):
...
+
+            nuclear_kwargs = args['postprocess_kwargs']['nuclear_kwargs']
+            assert nuclear_kwargs['radius'] == radius_nuc
+
+            # check that rest are unchanged
+            whole_cell_kwargs['maxima_threshold'] = 0.1
+            assert whole_cell_kwargs == default_cell_kwargs
+
+            nuclear_kwargs['radius'] = 2
+            assert nuclear_kwargs == default_nuc_kwargs
```
willgraf, line 257: Let's set this value based on the default_kwargs data, in case the defaults change in the future.

willgraf (changes requested):

willgraf (approved):
LGTM

PR #581, opened 2022-02-09
Modify testing workflow allow manual dispatch of workflow with different operating systems

msschwartz21:
## What
* Adds a workflow dispatch trigger to our testing workflow that allows a user to run the workflow with a different runner operating system
* The default operating system for tests triggered by a push or pull_request is ubuntu-latest

## Why
* Users have reported issues installing deepcell on particular operating systems. This modification to the workflow allows us to test operating systems as needed, but will not clutter our usage of github actions with testing all possible operating systems every time we run tests.

## Note
* Workflow_dispatch events usually don't work until they are merged into the default branch so although the manual dispatch is not visible now it should appear after merging.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -9,11 +9,12 @@ jobs:
 
   tests:
 
-    runs-on: ubuntu-latest
-
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


msschwartz21:
We are going to focus on providing support through docker containers and not to specific operating systems.

PR #584, opened 2022-03-04
Add option for graph attention layer to tracking model

msschwartz21:
## What
* Exposes an option to use the graph attention layer `GATConv` from `spektral`

## Why
* Allows us to test the impact of different graph layers on the tracking model

### DIFF
--- a/deepcell/model_zoo/tracking.py
+++ b/deepcell/model_zoo/tracking.py
@@ -217,6 +217,9 @@ class GNNTrackingModel(object):
         encoder_dim (int): Dimension of encoder
         embedding_dim (int): Dimension of embedding
         n_layers (int): number of layers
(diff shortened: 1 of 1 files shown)


deepcell/model_zoo/tracking.py
```diff
@@ -380,6 +380,8 @@ def get_neighborhood_encoder(self):
                 graph_layer = GCNConv(self.n_filters, activation=None, name=name)
             elif self.graph_layer == 'gcs':
                 graph_layer = GCSConv(self.n_filters, activation=None, name=name)
+            elif self.graph_layer == 'gat':
+                graph_layer = GATConv(self.n_filters, activation=None, name=name, attn_heads=4)
```
MekWarrior, line 384: should we expose the number of attention heads as well?

MekWarrior (approved):
Looks good, thank you for the changes.

PR #590, opened 2022-04-22
Pin `jinja2<3.1` to restore functional docs builds

msschwartz21:
## What
* Pins `jinja2<3.1` in `docs/requirements-docs.txt`

## Why
* The new `jinja2` release deprecates functions that are required for `sphinx`.


MekWarrior (approved):
It may be a good idea for the long haul to remove the dependency if possible since its deprecated.

PR #592, opened 2022-05-06
Update MultiplexSegmentation model

msschwartz21:
Update the mesmer model from model-registry#15. I'll leave this PR as a draft until @ngreenwald has a chance to test post-processing parameters and determine if any changes are needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,9 @@ ARG TF_VERSION=2.5.1-gpu
 
 FROM tensorflow/tensorflow:${TF_VERSION}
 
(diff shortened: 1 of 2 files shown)


ngreenwald:
Looks like the URL is private
image

msschwartz21:
Good catch! Thank you for checking the link! It should be good accessible now.

msschwartz21:
It looks like the tar file didn't grab all of the correct files for the saved model when I made it. I'll need to wait until madrox is back online, but I'll try taring it again and see if I can correct the error.

ngreenwald (approved):
Looks good. Just tried it out and everything looks great. I'm going to hold off on seeing if the post-processing params need to be tuned until the next version is merged in

PR #597, opened 2022-05-21
Dvv/tfrecord

vanvalen:
## What
Included functions to save datasets as tfrecords and load them into tf.data.Dataset objects

## Why
As our training datasets grow, it is becoming difficult to load full datasets into memory. By introducing support for tfrecords, we can load portions of datasets from disk on the fly during training.


deepcell/utils/tfrecord_utils.py
```diff
@@ -0,0 +1,220 @@
...
+# to endorse or promote products derived from this software without specific
+# prior written permission.
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""deepcell-toolbox Data Utilities"""
```
msschwartz21, line 26:
```suggestion
"""Functions for reading and writing segmentation datasets as TF Records"""
```

msschwartz21 (changes requested):
A few requests
- Add the import for tfrecord_utils to `deepcell/utils/__init__.py`
- Update the tests to return an integer image for y data

msschwartz21 (approved):

PR #606, opened 2022-06-04
Fix tracking model bug that pinned n_filters, encoder_dim and embedding_dim to 64

vanvalen:
## What
* Fixed a bug that required the tracking model to have n_filters, encoder_dim, and embedding_dim be pinned to 64

## Why
* Model optimization is going to require us to change these parameters to improve performance and reduce model size. This pull request makes that substantially easier by fixing this bug.

### DIFF
--- a/deepcell/data/tracking.py
+++ b/deepcell/data/tracking.py
@@ -70,8 +70,6 @@ class Track(object):  # pylint: disable=useless-object-inheritance
                 'Please make sure you are using a valid .trks file')
         self.appearance_dim = appearance_dim
         self.distance_threshold = distance_threshold
-        self.crop_mode = kwargs.get('crop_mode', 'resize')
-        self.norm = kwargs.get('norm', True)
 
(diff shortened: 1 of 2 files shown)


msschwartz21 (approved):
Can you leave a note in this PR thread about what new kwargs would be passed into deepcell.data.tracking.Track?

deepcell/data/tracking.py
```diff
@@ -53,7 +53,8 @@
 class Track(object):  # pylint: disable=useless-object-inheritance
 
     def __init__(self, path=None, tracked_data=None,
-                 appearance_dim=32, distance_threshold=64):
+                 appearance_dim=32, distance_threshold=64,
+                 **kwargs):
```
msschwartz21, line 57: Why are kwargs being exposed here? Is this coming from the other PR to expose crop mode and norm?

vanvalen: Yeah I think that's a hold over from the other PR.

PR #609, opened 2022-06-10
Update mesmer post-processing args

ngreenwald:
## What
Updated the post-processing parameters for the Mesmer model. Also updates the notebook to describe how post-processing can be modified. 

## Why
The newly retrained model has different parameters that give the best results. In addition, I've gotten questions from a few different people about how to tweak the model output, having it in the notebook will make it easy for people to see the effects.


msschwartz21 (changes requested):
The updates to the notebook look good! Two quick requests:
- Fix the typo in dictionary in the last paragraph in the notebook
- Change the link to the model defaults to point to the master branch instead of a specific commit: https://github.com/vanvalenlab/deepcell-tf/blob/master/deepcell/applications/mesmer.py#L272

msschwartz21 (approved):

PR #627, opened 2022-11-16
Bump model versions for nuclear segmentation and tracking

msschwartz21:
Updates the nuclear segmentation model from model-registry #34 and the tracking model from model-registry #36


elaubsch (approved):
LGTM!

PR #637, opened 2023-01-05
Update Dockerfile to jupyter lab

rdilip:
## What
* Removed jupyter notebook call

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,4 +32,4 @@ RUN pip install /opt/deepcell-tf
 # Copy over deepcell notebooks
 COPY notebooks/ /notebooks/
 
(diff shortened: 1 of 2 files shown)


msschwartz21 (dismissed):
I'm pretty sure we also need to update the requirements to include jupyterlab. Although our tests do use our docker image, we override the entrypoint, so the tests wouldn't catch that we are calling jupyter lab without installing it.

msschwartz21 (changes requested):
See comment above

rdilip:
Update: the image builds and runs without any issues.

msschwartz21 (approved):

PR #666, opened 2023-06-06
Update nuclear tracking and segmentation model paths

msschwartz21:
This PR updates the models used by the `NuclearSegmentation` and `CellTracking` applications to the latest versions associated with the upcoming paper.


msschwartz21:
@rossbar It looks like the test that is failing is testing a function marked as deprecated
https://github.com/vanvalenlab/deepcell-tf/blob/f3061d605f962a94e9832f0144180a19021cb69b/deepcell/utils/io_utils.py#L36-L37

Can I just delete the test function for `get_image`?

rossbar (approved):
LGTM, thanks @msschwartz21 !

PR #681, opened 2023-08-02
Application gallery and authenticated applications

msschwartz21:
Currently this is just the set up for a documentation gallery for our model applications

- [x] Use the fetch_data function to retrieve models (#678)
- [x] Update data access in scripts to use sample data objects once #679 is merged


rossbar (approved):
Awesome, another impressive feat - thanks @msschwartz21 

I went ahead and actually ran the applications examples locally in as many combinations as I could think of. I'm sure there are still a few remaining rough edges on the Python-side of the authentication system, but hopefully I've gotten some of the most obvious.

Everything else (docs gallery, etc.) LGTM!

PR #688, opened 2023-08-29
Authenticated SpotNet dataset

elaubsch:
This PR adds the `SpotNet` dataset to the list of new authenticated datasets, following the established pattern. It also adds a `SpotNetExampleData` class, which loads the example data for the [Polaris example notebooks](https://github.com/vanvalenlab/deepcell-spots/tree/master/notebooks). This class deviates slightly from the existing pattern for loading data in order to accommodate for the different file types that need to be loaded. The SpotNet dataset has also been added to the datasets gallery in the documentation.


rossbar (approved):
Thanks @elaubsch , this LGTM. The doc failures are expected and unrelated - the docs related to the SpotNet dataset look good!

PR #750, opened 2026-05-30
LIC: Revert licence text to exactly match OSI Apache 2

rossbar:
* Exact OSI Apache 2 text in `LICENSE`
* Remove per-file license notices; package license covers all software


rossbar:
@vanvalen please take a look and verify for the record. Once I have approval I'll go ahead and apply the same pattern to the other repos.

